PR #67, opened 2022-09-19
Fix/ut140 ut141 name filtering

olzzon:
This fix includes:
- Filtering of folders where a sub sub folder had the same name as another sub sub folder
- Double tally indication if 2 files in different folders had the same filename.
- Basic Refactoring of serverside CasparCGhandler


olzzon:
By accident i made a release. (as it's still using the old workflow)
We should make a new release when merged to develop

PR #68, opened 2022-09-20
feat: Stettings startupstate of Loop, Mix and Manual

olzzon:
Added startup state of loop, mix, manual in settings, for each output.


KvelaGorrrrnio (approved):
A suggestion is to consider using nullish coalescing (`??`) instead of the `||` in some of the places.

PR #69, opened 2022-11-29
Fix/cliptool settings lost

olzzon:
Fix: Prevent initial settings to be written to file. (could happen if storing before the settings had been reloaded)
Feat: Confirm a ClipTool restart (using window.confirm)
Feat: Hide restart when a client is on a single output. This would cause all outputs to stop, also the ones that the client didn't know about.

### DIFF
--- a/src/server/utils/SettingsStorage.ts
+++ b/src/server/utils/SettingsStorage.ts
@@ -19,7 +19,7 @@ export const loadSettings = () => {
 }
 
 export const saveSettings = () => {
-    let stringifiedSettings = JSON.stringify(reduxState.settings[0].generics)
+    const stringifiedSettings = JSON.stringify(reduxState.settings[0].generics)
     if (!fs.existsSync('storage')) {
         fs.mkdirSync('storage')
     }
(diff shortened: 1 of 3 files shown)


src/server/utils/SettingsStorage.ts
```diff
@@ -18,17 +19,23 @@ export const loadSettings = () => {
...
+        stringifiedSettings !==
+        JSON.stringify(defaultSettingsReducerState()[0].generics)
+    ) {
+        fs.writeFile(
+            path.resolve('storage', 'settings.json'),
+            stringifiedSettings,
+            'utf8',
+            (error) => {
+                console.log(error)
+            }
```
KvelaGorrrrnio, line 36: Can be simplified to this:
```suggestion
            (error) => console.log(error),

```
And perhaps warn or error would be better suited log level.

KvelaGorrrrnio (approved):
Seems fine.

src/server/utils/SettingsStorage.ts
olzzon: I didn't touch logging, because next plan is to update to mediatech logger.
There's still a lot of console.log() in the code.

PR #71, opened 2022-12-04
Feat/http html overlay

olzzon:
Support for externally generated overlay graphics


src/client/components/Header.tsx
```diff
@@ -67,12 +67,26 @@ const handleMixStatus = () => {
...
+}
+
 const mixStateStyle = () => {
     return reduxState.media[0].output[reduxState.appNav[0].activeTab]?.mixState
         ? { backgroundColor: 'rgb(28, 115, 165)' }
         : { backgroundColor: 'grey' }
 }
 
+const webStateStyle = () => {
+    return reduxState.media[0].output[reduxState.appNav[0].activeTab]?.webState
```
LindvedKrvang, line 85: I think this would make it a bit more readable:
```
const isWebpage: boolean = reduxState.media[0].output[reduxState.appNav[0].activeTab]?.webState
return { backgroundColor: isWebpage ? 'rgb(28, 115, 165)' : 'grey' }
```
Or you could do something like:
```
const backgroundColor: string = reduxState.media[0].output[reduxState.appNav[0].activeTab]?.webState ? 'rgb(28, 115, 165)' : 'grey'
return { backgroundColor }
```
I am most in favor for the first one.

The big take away here is that when have you these long chains: `reduxState.media[0].output[reduxState.appNav[0].activeTab]?.webState` it is a very good idea to store their value in an aptly named variable. This way it's way quicker to get an idea of what it is suppose to be and we don't have to guess. (I guessed that the variable was supposed to be whether or not we are on a webpage, but I'm fairly certain that is not correct. But reading this code snippet, then that is my best guess - and guesses leads to bugs because guesses can be wrong.)

Sidenote: You are dancing on a thin line with the Law of Demeter with these chains here.

src/client/components/Settings.tsx
```diff
@@ -201,6 +201,18 @@ const RenderOutputSettings = () => {
         reduxStore.dispatch(setGenerics(generics))
     }
 
+    const handleStartupWebState = (event) => {
```
LindvedKrvang, line 204: I would highly recommend that you don't use the `handleX` naming convention.
Reason for this is that `handleOkButtonClicked` doesn't really convey what the method does. It only conveys when the method is supposed to be used. This is a piece of information that the method itself should not know about since there is nothing stopping you from using the method `reloadPageOnButtonClicked()` in the response handler of a http request.

Given this code snippet:
```
<button onClick="() => handleConfirmClicked()">Confirm</button>
```
What happens when the button is clicked?

We simply don't know and are now force to go look at implementation to see if `handleConfirmClicked` is relevant for what we currently are doing.

But if the snippet was:
```
<button onClick="() => sendDeleteUserRequest()">Confirm</button>
```
We now have a pretty good idea of what happens when the button is clicked. And if what we currently are doing has nothing to do with the deletion of a user, we can then "disregard" the method without ever looking at it, which saves us time when debugging/maintaining the software.


Sidenote: If you keep using the `handleX` naming convention, what will do if you have two OK buttons or two Confirm buttons on the same page that those different things? Will you start naming them `handleOkButtonClicked` and `handleOkButtonClicked2` - how do you know which method belongs to which button?

src/server/handlers/CasparCgHandler.ts
```diff
@@ -140,12 +142,30 @@ const dispatchConfig = (config: any) => {
...
+            )
+        )
     })
     logger.info('Number of Channels : ' + config.channels.length)
     socketServer.emit(IO.SETTINGS_UPDATE, reduxState.settings[0])
     initializeClient()
 }
 
+const loadInitalOverlay = () => {
+    if (reduxState.settings[0].generics.startupWebState) {
```
LindvedKrvang, line 158: To avoid unnecessary nesting it's a good practice to invert you if-statements.
Then it would look something like this:
```
const loadInitalOverlay = () => {
  if (!reduxState.settings[0].generics.startupWebState) {
    return
  }
  reduxState.settings[0].ccgConfig.channels.forEach((ch, index) => {
        playOverlay(
            index,
            10,
            reduxState.settings[0].generics.webURL[index]
        )
   })
}
```

The difference might seem small on this method but when you start having multiple layers and nested if-statementns, this is going to make a world of a difference

LindvedKrvang (approved):

src/server/handlers/CasparCgHandler.ts
olzzon: Thanks, good point.

src/client/components/Header.tsx
olzzon: Over different types of code, i always prefer consistency in the code. So if all mixStateStyle, loopStateStyle etc. are written this way, I think this also should be done like that. At least until a refactor.
Performance wise i also think that the current implementation is better.

src/client/components/Settings.tsx
olzzon: I see your point, not sure i completely agree though, as handleX is one of the most used syntax' in React when handling events.
Besides that over different types of code, i always prefer consistency in the code. So if all mixStateStyle, loopStateStyle etc. are written this way, I think this also should be done like that. At least until a refactor.

PR #76, opened 2023-01-11
fix: Updated incorrect log statements and made logs more consistent.

KvelaGorrrrnio:
Updated some log statements to correct syntax along with formatted the message.
Removed unused import statements and unused method.

### DIFF
--- a/src/gateway/gateways/AmpServerGateway.ts
+++ b/src/gateway/gateways/AmpServerGateway.ts
@@ -38,6 +38,32 @@ export const ampServerGateway = () => {
     })
 }
 
+// This is preparation for handling other commands in the AMP protocol.
+// TODO: Make use of it.
+function checkAmpCommand(
+    message: string,
+    command: string | undefined
+): boolean {
+    if (!command || !message) return false
+    if (message === command) return true
+    let messageArray: string[] = message.split('/')
+    let commandArray: string[] = command.split('/')
+    let status: boolean = true
+    if (messageArray.length !== commandArray.length) {
+        return false
+    }
+    commandArray.forEach((commandPart: string, index: number) => {
+        if (commandPart === '{output}') {
+            if (typeof parseFloat(messageArray[index]) !== 'number') {
+                status = false
+            }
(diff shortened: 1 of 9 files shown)


src/gateway/gateways/AmpServerGateway.ts
```diff
@@ -35,34 +34,10 @@ export const ampServerGateway = () => {
     }).listen(3811)
     let ipAddresses = getThisMachineIpAddresses()
     ipAddresses.forEach((address) => {
-        logger.info(`AMP Host Listening for TCP at: ${address}, Port: 3811`)
+        logger.info(`AMP Host Listening for TCP at ${address}:3811.`)
     })
 }
 
-const checkAmpCommand = (
```
olzzon, line 42: The checkAmCommand was a preparation for handling other command in the AMP protocol.
Instead of removing it, IMHO we should consider keeping with a comment of the purpose.

olzzon (approved):

RasmusAlbrektsen (approved):

src/gateway/gateways/AmpServerGateway.ts
KvelaGorrrrnio: I will re-add it.

src/gateway/index.ts
```diff
@@ -3,11 +3,10 @@ import { logger } from './util/loggerGateway'
 
 import { reduxState } from '../model/reducers/store'
 import { app } from './app'
-import { socket } from './util/SocketGatewayHandlers'
 
-logger.info(' Runtime Arguments : %O', ARG_CONSTANTS)
+logger.data(ARG_CONSTANTS).info(' Runtime Arguments:')
```
LindvedKrvang, line 7: Is there a reason for the leading space?

LindvedKrvang (approved):

src/gateway/index.ts
KvelaGorrrrnio: I don't think so. I have removed the leading space.

PR #77, opened 2023-02-03
Ut 205 - Fixes Incorrect Thumbnail being shown

andr9528:
Note the changes in 'expressHandler.ts', might effect running cliptool for development on mac or from the published version.
Either case hasn't been tests - yet. Will check if the published version works as intended, ones we are back from the next few days of Design Thinking. Would be happy if @KvelaGorrrrnio could check if it runs as intended on mac, if it is even suppose to be able to do that.

PR also includes abit of boyscouting i did when i looked around the code.


src/client/components/Header.tsx
```diff
@@ -16,6 +16,9 @@ import * as IO from '../../model/SocketIoConstants'
 import { findThumbPix } from './Thumbnail'
 import { SettingsPage } from './Settings'
 
+const offColor = { backgroundColor: 'grey' }
```
LindvedKrvang, line 19: It would be nice that static constants variables like this used SNAKE_CASE.
This makes it far easier to spot that they are not local variables when reading a method that uses these variables.

Minor thing: Not sure if this should be changed, but something bugs me about that it's just being called `...Color` when it's actually a CSS attribute/value pair.

Minor minor thing: Is `on` and `off` the most fitting words? (I don't know what this is used for, so maybe they are) Are the things used by this actually being turned on and off or there they, for example, being selected/deselected?

src/client/components/Thumbnail.tsx
```diff
@@ -14,7 +14,7 @@ export const findThumbPix = (fileName: string, channelIndex: number) => {
     let thumb =
         reduxState.media[0].output[channelIndex]?.thumbnailList.filter(
             (item: IThumbFile) => {
-                return fileName.toUpperCase().includes(item.name)
+                return item.name.toUpperCase().match(fileName.toUpperCase())
```
LindvedKrvang, line 17: Curious: What's the reason for changing the order of `fileName` and `item`?

Also, this could potentially throw an undefined error if `thumbnailList` is allowed to have undefined entries (I hope it's not allowed)

src/client/components/Thumbnail.tsx
```diff
@@ -77,14 +77,9 @@ export const Thumbnail = () => {
...
 const handleClickMedia = (fileName: string) => {
-    if (
-        !reduxState.media[0].output[reduxState.appNav[0].activeTab]
-            ?.manualstartState
-    ) {
-        socket.emit(PGM_PLAY, reduxState.appNav[0].activeTab, fileName)
-    } else {
-        socket.emit(PGM_LOAD, reduxState.appNav[0].activeTab, fileName)
-    }
+    const event = !reduxState.media[0].output[reduxState.appNav[0].activeTab]
```
LindvedKrvang, line 80: Personal preference here: A way of formatting ternaries that I have found really helpful and easy to read is this:
```
const event = !reduxState.media[0].output[reduxState.appNav[0].activeTab]?.manualstartState 
    ? PGM_PLAY 
    : PGM_LOAD
```
Then you have condition on one line. Case 1 on another and case 2 on a third. And by putting the operators as the first thing you don't have to scrutinize a potentially long conditional (as is the case here) to find out that it's in fact a ternary.

.gitignore
```diff
@@ -12,7 +12,6 @@ release-builds/
 build/
 node_modules/
 *.compiled.*
-yarn.lock
```
LindvedKrvang, line 15: Do we know why this was added to the gitignore to begin with? It seems counter-intuiative

LindvedKrvang (approved):

src/client/components/Thumbnail.tsx
olzzon: Be aware that filename and item.name is not always the same, it's dependent on whether it's received through the AMCP or OSC protocol. . So it should be the fileName that are returned and not the item.name.

src/server/handlers/expressHandler.ts
```diff
@@ -13,7 +13,9 @@ app.use('/', express.static(path.join(__dirname, '../../client')))
 server.on('connection', () => {
     app.get('/', (req: any, res: any) => {
         logger.info('Connected Client')
-        res.sendFile('index.html')
+        res.sendFile('index.html', {
+            root: path.join(__dirname, '../../client/src/client'),
+        })
```
olzzon, line 18: If the "cp"/"copyfile" works in package.json, then res.sendFile('index.html') will work.
I don't think that we should link to a path in the source code.

olzzon (commented):
In general, i don't agree on the concept of fixing a bug, and refactoring in the same release.
And as there's a dedicated bug fix. I think we should make a release with only the fix. And then do refactoring afterward.
The most important thing in a refactor would be to normalize the AMCP and the OSC responses, including differences that could ocour between Windows and Linux

src/client/components/Header.tsx
andr9528: The On/Off color is used for the style of 4 buttons, where the color indicates if they are on or off, thus naming them On/Off, fits well - atleast in my mind. Feel free to come up with a better suggestion.

Have applied your suggestion of using SNAKE_CASE.

src/client/components/Thumbnail.tsx
andr9528: @LindvedKrvang 
My reasoning for flipping the order was mostly because of how i read it. 
When you have the `fileName` first, it is the 'gate' asking if something is allowed to pass, instead of the 'arrival' asking if it is allowed to pass. Thus i like to read it as many (arrivals) askes one (gate), instead of one askes many. Propably not a good reasoning but meh `¯\_(ツ)_/¯`.

@olzzon 
It is not the `fileName` that is returned, but the boolean value gained from matching `fileName` and `item.name`.
Have however flipped the order of `item.name` and `fileName` back, not that i think it should have any effect.
Does doing `stringA.match(stringB)` and `stringB.match(stringA)`, not do exactly the same in alle cases?

src/client/components/Thumbnail.tsx
andr9528: I also like that styling, and my Resharper does that automatically for most ternaries lines when doing C#.
However my Resharper has not been configured for Typescript, so i just forgot to do it manually.
Have done so now.

src/client/components/Thumbnail.tsx
olzzon: @andr9528 sorry my fault 😊

src/server/handlers/expressHandler.ts
andr9528: If it does not have the `root` part then i were unable to access Cliptool, when running in development.
From looking into it, it appeared that it was trying to open the file at a different location then before.

For the record, i do not like the `root` section either, but to get it to work i needed it. 
If you can give me a better solution to it being unable to locate `index.html`, when it doesn't have the `root` part, then i am all ears.

@KvelaGorrrrnio
Perhabs how files are copied are different with `copyfile` compared to how it was before with `cp`?

andr9528:
@olzzon 
> In general, i don't agree on the concept of fixing a bug, and refactoring in the same release. And as there's a dedicated bug fix. I think we should make a release with only the fix. And then do refactoring afterward. The most important thing in a refactor would be to normalize the AMCP and the OSC responses, including differences that could ocour between Windows and Linux

Do you know about the concept of boy scouting?
It is the concept of leaving the code better than you found it. 

Of cource you can't refactor all of the code and call it boy scouting, but smaller changes here and there, to ultimately increase the code quality, is the general expected outcome.

In the case of this PR, i have made 3 changes that i classify as boy scouting, as they don't do any major reworking of how something works, but simply improves the readability or lowers duplication in the code.
- I removed an unused line, that Anders and I spotted.
- I extracted the same color stylings to shared constants, reducing duplication.
- I simplified an event emit, by extracting the choice of event being fired to a ternary, reducing duplication.

If i'd chosen to restructure e.g Thumpnail.tsx, such that the stylings were in their own file, and each subcomponent in the file, had thier own file, then yes that would have NOT been an acceptably change, when tasked with doing a small bugfix.

Hope you get what i mean. Refactors can come in small and large sizes. It is ok to do a few small ones, together with the main task. Doing a larger refactor, like the one you mentioned, requires a dedicated task for said refactor.

src/client/components/Thumbnail.tsx
andr9528: Just FYI, the order DOES matter. The way i did fixes the bug, BUT if i flip the order back, even when using `.match(...)`, the bug is back.

Not entirely sure why the order has an effect. `¯\_(ツ)_/¯`

Disregard me saying anything about me flipping it back, in my previous message.

olzzon:
@andr9528 Yes, I'm completely aware of the concept, and that the team are using it, so it's a general concern,.
In my view it's perfect for new releases, but not for fixing a bug in a production critical environment.

And though the changes you have are great and an improvement, i do think that the bugfix should be isolated in a p.r.
This way 100% the focus would be on whether it's fixing the problem, or introducing other problems.


Btw. In this P.R there's also 3 other changes beside the ones mentioned, and here's one that could have the risk of causing problems in the build process:
```
      res.sendFile('index.html', {
            root: path.join(__dirname, '../../client/src/client'),
        })
```

.gitignore
andr9528: Anders also questioned why that had been added to the gitignore. 
He was the one to say that the line should be deleted from the gitignore. 

Had he not said it, i might still have deleted it myself, to make this repo match how we do it in the other repos.

RasmusAlbrektsen (approved):
Looks good and i like the recent changes as well. Of course this is dependent on testing and ensuring whether or not the path added to `res.sendFile()` is ideal/unavoidable in this situation.

andr9528:
Just did a local test of a packaged version of Cliptool, started via the CasperCG Launcher, and from what i see the `res.sendFile(...)` still works as intended, along with the thumbnail showing the correct image.

Only needs to test that i doesn't break something in production environment. From my tests it shouldn't.

olzzon (approved):

PR #78, opened 2023-02-10
UT-206: Adding soft-delete of thumbnails

andr9528:
Created as a draft PR, so developers (mainly Kasper) can come with feedback on changes prior to full on PR review.

PR reviewed for: 
- [x] UT-206
  - [x] UT-207
  - [x] UT-208
  - [x] UT-209


src/model/reducers/mediaReducer.ts
```diff
@@ -66,58 +69,101 @@ export const media = (state: Array<IMedia> = defaultMediaState(), action) => {
...
+                    (nextState[0].output[action.channelIndex].hideState =
+                        action.hideState)
+            )
+            break
     }
+
+    return nextState
+}
+
+function conditionalApplyAction(
```
LindvedKrvang, line 162: This is overengineered.
I see what you are trying to achieve, but it's enough to just extract the conditional:
``` Typescript
function doesChannelExist(state: StateType, action: ActionType): boolean { // or whatever the conditional is trying to find out
  return state[0].output.length >= action.channelIndex
}
```

So now your switch cases would just look like this:
``` typescript
case: IO.SET_TIME:
  if (doesChannelExist(nextState, action)) {
    nextState[0].output[action.channelIndex].time = action.time
  }
  return nextState
}
```

src/model/SocketIoConstants.ts
```diff
@@ -5,7 +5,8 @@ export const THUMB_UPDATE = 'thumbUpdate'
 export const FOLDERS_UPDATE = 'foldersUpdate'
 export const TAB_DATA_UPDATE = 'tabDataUpdate'
 export const SETTINGS_UPDATE = 'settingsUpdate'
-export const LOOP_STATEUPDATE = 'loopStateUpdate'
+export const LOOP_STATE_UPDATE = 'loopStateUpdate'
+export const HIDE_STATE_UPDATE = 'hideStateUpdate'
```
LindvedKrvang, line 9: `HIDE_STATE` is a weird phrasing especially when the english language offers words that describe this very thing. (`HIDDEN` or `VISIBILITY` - I vote for visibility; then it would be `UPDATE_VISIBILITY` which is more like normal english).

src/model/reducers/mediaActions.ts
```diff
@@ -10,6 +10,7 @@ export const SET_MIX = 'setMix'
 export const SET_WEB = 'setWeb'
 export const SET_MANUAL_START = 'setManualStart'
 export const SET_TIME = 'setTime'
+export const SET_HIDE = 'setHide'
```
LindvedKrvang, line 13: If you go with `VISIBILITY` instead of `HIDE` then this would be `SET_VISIBILITY` which would make more sense since `hide` is a verb and you don't set a verb, you do a verb.

src/model/reducers/mediaActions.ts
```diff
@@ -97,3 +98,11 @@ export const setTime = (channelIndex: number, time: [number, number]) => {
         time: time,
     }
 }
+
+export const setHide = (channelIndex: number, hideState: boolean) => {
```
LindvedKrvang, line 102: Continuing the visibility path, this would be:
``` typescript
export const setVisibility = (channelIndex: number, isVisible: boolean) => {
  return {
    type: SET_VISIBILITY,
    channelIndex,
    visibility: isVisible
  }
}
```

src/model/reducers/mediaReducer.ts
andr9528: Noted, and Applied.

src/client/components/Header.tsx
```diff
@@ -98,6 +98,21 @@ const handleManualStartStatus = () => {
...
+function handleIsHidingState() {
+    socket.emit(
+        IO.SET_HIDE_STATE, 
+        reduxState.appNav[0].activeTab, 
+        !reduxState.media[0].output[reduxState.appNav[0].activeTab]
+            .hideState
+    )
+}
+
+function isHidingStyle() {
```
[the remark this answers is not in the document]
andr9528: Name updated to not include 'is', match new 'visibility' naming, and type added.

src/model/SocketIoConstants.ts
andr9528: Updated to use 'VISIBILITY' instead of 'HIDE'.

src/model/reducers/mediaActions.ts
andr9528: Updated

src/model/reducers/mediaActions.ts
andr9528: Updated

src/client/components/Thumbnail.tsx
```diff
@@ -42,58 +40,67 @@ export const isThumbWithTally = (thumbName): boolean => {
...
+        )    
+}
+
+const handleClickMedia = (fileName: string) => {    
+    const isTogglingVisibility = reduxState.media[0].output[reduxState.appNav[0].activeTab]?.visibilityState
+    const onClickAction = isTogglingVisibility 
+        ? onClickToggleVisibility 
+        : onClickPlayClicked
+
+    onClickAction(fileName)
```
KvelaGorrrrnio, line 78: By using the visibilityState as a boolean, this function will be hard to extend with other modes. I refer to my previous comment on using an enum of operations mode. By doing so we enable easy extension of new operations modes for interacting with the thumbnails.

```typescript
function handleClickMedia(fileName: string): void {
  const operationsMode = reduxState.media[0].output[reduxState.appNav[0].activeTab].operationsMode

  switch (operationsMode) {
    case OperationsMode.EDIT_VISIBILITY:
      onClickToggleVisibility(fileName)
      break

    case OperationsMode.CONTROL:
    default:
      onClickPlayClicked(fileName)
  }
}
```

src/client/components/Thumbnail.tsx
```diff
@@ -42,58 +40,67 @@ export const isThumbWithTally = (thumbName): boolean => {
...
 }
 
-const handleClickMedia = (fileName: string) => {
+function onClickToggleVisibility(fileName: string) {
```
KvelaGorrrrnio, line 81: The function might be used outside of the click flow. It could be called `toggleVisibility`.

src/client/components/Thumbnail.tsx
```diff
@@ -42,58 +40,67 @@ export const isThumbWithTally = (thumbName): boolean => {
...
+    socket.emit(TOGGLE_THUMBNAIL_VISIBILITY, reduxState.appNav[0].activeTab, fileName)
+}
+
+function onClickPlayClicked(fileName: string ) {
```
KvelaGorrrrnio, line 85: The function might be used outside of the click flow. It could be called `playClip` or something similiar.

src/client/components/Thumbnail.tsx
```diff
@@ -114,52 +121,64 @@ const RenderThumb = (props) => {
...
-                      ]?.videoFormat?.frameRate
-                  )
+                ? secondsToTimeCode(time, frameRate)
                 : ''}
         </a>
     )
 }
 
-const RenderThumbPix = (props) => {
+const RenderThumbPix = (props: {item: IMediaFile}) => {
```
KvelaGorrrrnio, line 143: The prop `item` is of type `IMediaFile` and would semantically benefit from being renamed to `file`.

src/client/components/Thumbnail.tsx
```diff
@@ -114,52 +121,64 @@ const RenderThumb = (props) => {
...
-                isThumbWithTally(props.item.name)
-                    ? { borderWidth: '4px' }
-                    : { borderWidth: '0px' }
-            )}
+            style={{
+                ...borderStyle(props.item.name),
+                filter: props.item.name in hiddenFiles
+                    ? 'grayscale(1)' 
+                    : 'grayscale(0)'
+            }}
```
KvelaGorrrrnio, line 167: As we talked about, this should simply set a className, which then is responsible for styling the component.

src/client/components/Thumbnail.tsx
```diff
@@ -114,52 +121,64 @@ const RenderThumb = (props) => {
...
 
-const RenderThumbText = (props) => {
+function borderStyle(filepath: string) {
+    return Object.assign(
+        {},
+        isThumbWithTally(filepath)
+            ? { borderWidth: '4px' }
+            : { borderWidth: '0px' }
+    )
+}
```
KvelaGorrrrnio, line 179: I know that this is simply a refactor, but I would suggest to also set a class name here and then use CSS for styling.

src/client/util/SocketClientHandlers.ts
```diff
@@ -52,6 +58,13 @@ socket.on(IO.THUMB_UPDATE, (channelIndex: number, payload: IThumbFile[]) => {
     reduxStore.dispatch(updateThumbFileList(channelIndex, payload))
 })
 
+socket.on(
+    IO.HIDDEN_FILES_UPDATE,
+    (channelIndex: number, hiddenFiles: Record<string, IChangedInfo>) => {
```
KvelaGorrrrnio, line 63: `channelIndex` should be removed since it is not being used.

src/gateway/util/SocketGatewayHandlers.ts
```diff
@@ -82,11 +83,16 @@ socket.on(IO.TIME_TALLY_UPDATE, (data: IO.ITimeTallyPayload[]) => {
...
 
-socket.on(IO.LOOP_STATEUPDATE, (channelIndex: number, loop: boolean) => {
+socket.on(IO.LOOP_STATE_UPDATE, (channelIndex: number, loop: boolean) => {
     reduxStore.dispatch(setLoop(channelIndex, loop))
     console.log('Loop State updated')
 })
 
+socket.on(IO.VISIBILITY_STATE_UPDATE, (channelIndex: number, hide: boolean) => {
+    reduxStore.dispatch(setVisibility(channelIndex, hide))
+    console.log('Visibility State updated')
```
KvelaGorrrrnio, line 93: Is this a development log?

src/model/reducers/mediaActions.ts
```diff
@@ -1,15 +1,17 @@
-import { IMediaFile, IThumbFile } from './mediaReducer'
+import { IChangedInfo, IMediaFile, IThumbFile } from './mediaReducer'
```
KvelaGorrrrnio, line 1: As we talked about, `IChangedInfo` would benefit from getting a more descriptive name.

src/server/handlers/socketIOServerHandler.ts
```diff
@@ -17,12 +17,20 @@ import {
...
 import { saveSettings } from '../utils/SettingsStorage'
-import { IOutput } from '../../model/reducers/mediaReducer'
+import {
+    IChangedInfo,
+    IMedia,
+    IMediaFile,
+    IOutput,
+} from '../../model/reducers/mediaReducer'
 import { assignThumbNailListToOutputs } from './CasparCgHandler'
+import { saveHiddenFiles } from '../utils/HiddenFilesStorage'
```
KvelaGorrrrnio, line 33: `HiddenFilesStorage` should be renamed to `hiddenFilesStorage`.

KvelaGorrrrnio (commented):
A lot of the comments are stuff we already talked about today :)

src/gateway/util/SocketGatewayHandlers.ts
andr9528: No. I replicated the same layout for the event handler as the other event handlers in this file. 
They all have a `console.log(...)` after their `.dispatch(...)` so i added one too.

src/server/handlers/socketIOServerHandler.ts
andr9528: The file that does the loading/saving of settings was spelled with capital start letter, thus i also did it here. Don't know which is more correct.

Will rename it to non-capital start letter. Let me know, if i should do the same for 'SettingsStorage.ts'.

src/model/reducers/mediaActions.ts
andr9528: Don't remember us talking about this one?
What would you suggest it being named?

src/client/util/SocketClientHandlers.ts
andr9528: Gone

src/client/components/Thumbnail.tsx
```diff
@@ -42,58 +40,67 @@ export const isThumbWithTally = (thumbName): boolean => {
...
         (storeUpdate: any) =>
             storeUpdate.media[0].output[reduxState.appNav[0].activeTab]
                 ?.mediaFiles
     )
-
-    // Render:
-    if (reduxState.appNav[0].selectView === 0) {
+    const hiddenFiles: Record<string, IChangedInfo> = useSelector(
+        (storeUpdate: any) => 
+            storeUpdate.media[0].hiddenFiles
```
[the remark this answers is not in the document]
andr9528: Will try, but if the styling applied on commit is the one to do it, then i might not be able.

src/client/components/Thumbnail.tsx
andr9528: Renamed to mentioned name.

src/client/components/Thumbnail.tsx
andr9528: Renamed to 'playFile'

src/client/components/Thumbnail.tsx
andr9528: Updated to use a switch as described.

src/model/reducers/mediaActions.ts
andr9528: Renamed to IHiddenFileInfo.

src/client/components/Header.tsx
```diff
@@ -149,6 +164,15 @@ export const RenderFullHeader = () => {
                             OVERLAY
                         </button>
                     </div>
+                    <div className="App-button-background">
+                        <button
+                            className="App-switch-button"
+                            onClick={() => handleVisibilityState()}
```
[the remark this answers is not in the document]
andr9528: Turns out No. Just copied the layout of the other buttons. Will update them too.

src/client/components/Thumbnail.tsx
andr9528: Styling has now been extracted to CSS.

src/client/components/Thumbnail.tsx
andr9528: Styling has now been extracted to CSS.

src/client/components/Header.tsx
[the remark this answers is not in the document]
LindvedKrvang: Sidenote: If you had needed to parse in a parameter to the function, then the arrow funtion would have been necessary :)

src/client/components/Settings.tsx
```diff
@@ -27,6 +31,23 @@ const selectorColorStyles = {
...
+        IO.SET_OPERATION_MODE, 
+        reduxState.appNav[0].activeTab, 
+        reduxState.media[0].output[reduxState.appNav[0].activeTab]
+            .operationMode !== OperationMode.EDIT_VISIBILITY 
+            ? OperationMode.EDIT_VISIBILITY 
+            : OperationMode.CONTROL
+    )
+}
+
+function editVisibilityStyle(): { backgroundColor: string } {
```
LindvedKrvang, line 45: Your naming convention clashes with what you are doing. `editX` indicate that something will happen/change X. That's not what you are doing here. You are simply getting an object (and said object knows nothing about VisibilityStyle)

What you could do is refactor it to:
```
function getOperationMode(): boolean {
  return reduxState.media[0].output[reduxState.appNav[0].activeTab]?.operationMode
}
```
And then in your HTML:
```
<button 
    className="save-button { getOperationMode() === OperationMode.EDIT_VISIBILITY ? 'name-of-edit-visibility-css-class' : 'name-of-control-css-class'} "
    onClick={handleToggleEditVisibilityMode}>
```

src/client/components/Thumbnail.tsx
```diff
@@ -42,176 +44,195 @@ export const isThumbWithTally = (thumbName): boolean => {
...
+                {props.file.name
+                    .substring(props.file.name.lastIndexOf('/') + 1)
                     .slice(-45)}
             </p>
         </div>
     )
 }
 
-const RenderThumbTimeCode = (props) => {
+const RenderThumbTimeCode = (props: ThumbnailProps) => {
```
LindvedKrvang, line 141: This is a component. We should only have one component per file

src/client/components/Thumbnail.tsx
```diff
@@ -42,176 +44,195 @@ export const isThumbWithTally = (thumbName): boolean => {
...
-                  )
+            {isThumbWithTally(props.file.name)
+                ? secondsToTimeCode(time, frameRate)
                 : ''}
         </a>
     )
 }
 
-const RenderThumbPix = (props) => {
+const RenderThumbPix = (props: ThumbnailProps) => {
```
LindvedKrvang, line 160: This is a component. We should only have one component per file

src/client/components/Thumbnail.tsx
```diff
@@ -42,176 +44,195 @@ export const isThumbWithTally = (thumbName): boolean => {
...
-    ? PGM_PLAY 
-    : PGM_LOAD
-    socket.emit(event, reduxState.appNav[0].activeTab, fileName)    
+        ? PGM_PLAY 
+        : PGM_LOAD
+    socket.emit(event, reduxState.appNav[0].activeTab, fileName)
 }
 
-const RenderThumb = (props) => {
+const RenderThumb = (props: ThumbnailProps) => {
```
LindvedKrvang, line 101: This is a component. We should only have one component per file

src/client/components/Thumbnail.tsx
```diff
@@ -42,176 +44,195 @@ export const isThumbWithTally = (thumbName): boolean => {
...
-                    ? { borderWidth: '4px' }
-                    : { borderWidth: '0px' }
-            )}
-        />
+        <img src={url} className={classNames} />
     )
 }
 
-const RenderThumbText = (props) => {
+const RenderThumbText = (props: ThumbnailProps) => {
```
LindvedKrvang, line 187: This is a component. We should only have one component per file

src/client/components/Thumbnail.tsx
```diff
@@ -42,176 +44,195 @@ export const isThumbWithTally = (thumbName): boolean => {
...
+                {props.file.name
+                    .substring(props.file.name.lastIndexOf('/') + 1)
                     .slice(-45)}
             </p>
         </div>
     )
 }
 
-const RenderThumbTextTimeCode = (props) => {
+const RenderThumbTextTimeCode = (props: ThumbnailProps) => {
```
LindvedKrvang, line 221: This is a component. We should only have one component per file

src/client/css/Footer.css
```diff
@@ -0,0 +1,22 @@
...
+    margin-top: 20px;
+}
+
+.Footer-text {
+    color: orangered;
+    font-size: 35px;
+    text-align: center;
+}
+
+.Footer-button {
```
LindvedKrvang, line 12: CSS classes are all lower cases, so this should be `.footer-button`.

Also, you are making with custom/specific styling for this component. We should aim to have general styling used throughout the codebase. I.e. one `button-style` that is used for all buttons, so if/when we want to change the style of buttons we only have to do it once instead of potential hundreds of places

src/model/reducers/mediaActions.ts
```diff
@@ -1,15 +1,22 @@
...
 export const UPDATE_FOLDER_LIST = 'updateFolderList'
+export const UPDATE_HIDDEN_FILES = 'updateHiddenFiles'
 export const SET_TALLY_FILE_NAME = 'setTallyFileName'
 export const SET_NUMBER_OF_OUTPUTS = 'setNumberOfOutputs'
 export const SET_LOOP = 'setLoop'
 export const SET_MIX = 'setMix'
 export const SET_WEB = 'setWeb'
 export const SET_MANUAL_START = 'setManualStart'
 export const SET_TIME = 'setTime'
+export const SET_OPERATION_MODE = 'setVisibility'
```
LindvedKrvang, line 19: `setVisibility` does not match the name `SET_OPERATION_MODE`

src/model/reducers/mediaActions.ts
```diff
@@ -40,6 +47,15 @@ export const updateThumbFileList = (
     }
 }
 
+export function updateHiddenFiles(
```
LindvedKrvang, line 50: `updateX` indicates something is changing on X. Here you just wrap it in an object and returns it making `updateHiddenFiles` an inaccurate name for the function.

src/model/reducers/mediaReducer.ts
```diff
@@ -66,58 +77,71 @@ export const media = (state: Array<IMedia> = defaultMediaState(), action) => {
...
-            return nextState
+            break
         case IO.UPDATE_MEDIA_FILES:
-            if (nextState[0].output.length >= action.channelIndex) {
+            if (doesChannelExist(nextState, action)) {
                 nextState[0].output[action.channelIndex].mediaFiles =
                     action.files
             }
-            return nextState
+            break
```
LindvedKrvang, line 86: What was wrong with just returning immediatly?

src/model/reducers/mediaReducer.ts
```diff
@@ -66,58 +77,71 @@ export const media = (state: Array<IMedia> = defaultMediaState(), action) => {
...
+                nextState[0].output[action.channelIndex].operationMode =
+                    action.operationMode
+            }
+            break
     }
+
+    return nextState
+}
+
+function doesChannelExist(nextState: any, action: any) {
```
LindvedKrvang, line 145: Don't use `any` type

src/model/reducers/settingsReducer.ts
```diff
@@ -128,10 +140,10 @@ export const settings = (
                 ...action.channels.map(updateChannelConfigWithVideoFormat),
             ]
             nextState[0].ccgConfig.path = action.path
-            return nextState
+            break
```
LindvedKrvang, line 143: What was wrong with returning immediatly?

src/server/handlers/socketIOServerHandler.ts
```diff
@@ -103,6 +143,45 @@ export function socketIoHandlers(socket: any) {
         })
 }
 
+function toggleHiddenFile(
```
LindvedKrvang, line 146: This function is called `toggle`, but are you both CREATING and DELETING which is something completely different than toggling

src/server/utils/ccgHandlerUtils.ts
```diff
@@ -64,11 +64,13 @@ export const hasThumbListChanged = (
 ): boolean => {
     if (newList.length !== previousList.length) {
         return true
-    } else {
-        for (let i = 0; i < newList.length; i++) {
-            if (newList[i].name !== previousList[i].name) {
-                return true
-            }
+    }
+    for (let i = 0; i < newList.length; i++) {
```
LindvedKrvang, line 68:
```
export const hasThumbListChanged = (...): boolean => {
   ...
 
  return newList.some((entry, index) => entry.name !== previousList[i].size || entry.name !== previousList[i].size)
}
```

.gitignore
```diff
@@ -56,3 +56,4 @@ ehthumbs.db
 # Windows shortcuts #
 #####################
 *.lnk
+/storage/hiddenFiles.json
```
LindvedKrvang, line 59: What does this JSON do?

src/client/components/Footer.tsx
```diff
@@ -0,0 +1,48 @@
+import React from 'react'
+import { useSelector } from 'react-redux'
+import { OperationMode } from '../../model/reducers/mediaReducer'
+import { reduxState } from '../../model/reducers/store'
+import * as IO from '../../model/SocketIoConstants'
+
+import '../css/Footer.css'
```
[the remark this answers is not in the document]
andr9528: Didn't work without the import. Also how it is setup in the other files, so it makes sense for it to be needed.

src/client/components/Footer.tsx
```diff
@@ -0,0 +1,48 @@
...
+import { socket } from '../util/SocketClientHandlers'
+
+const footerDescriptions = new Map<OperationMode, string>([
+  [OperationMode.EDIT_VISIBILITY, 'You are currently editing visibility of files.'],
```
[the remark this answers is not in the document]
andr9528: I'd not exactly call it translations. It is just how i chose to keep the different strings, for the description and button text, with their Operation Mode key, in a manner that would be easily expandable, should the need arise.

src/client/components/Footer.tsx
```diff
@@ -0,0 +1,48 @@
...
+    OperationMode.CONTROL
+  )
+}
+
+export function Footer() {
+  const operationMode: OperationMode = useSelector((storeUpdate: any) =>
+    storeUpdate.media[0].output[reduxState.appNav[0].activeTab]?.operationMode)
+  
+  return (
+    <footer className='Footer'>
```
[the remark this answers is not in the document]
andr9528: Anders pointed this out too, to which i pointed out that many other CSS classes are spelled with a capital letter in this application, at the moment.

I would very much like to after this PR, to go through and update the code styling to be unified in how it is defined, along with a few other things that Anders and I have talked with Kasper about.
I am fairly certain i know how we, as a team, want the style to be, but i would be happy to have a chat about it, if you'd like.

src/client/components/Footer.tsx
```diff
@@ -0,0 +1,48 @@
...
+      { operationMode !== OperationMode.CONTROL 
+      ? <div className='Footer-flex'>
+          <div className='Footer-text'>
+            {footerDescriptions.has(operationMode) 
+            ? footerDescriptions.get(operationMode).toUpperCase()
```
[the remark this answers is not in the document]
andr9528: I have extracted the innards of the Footer to a seperate functional component, in this file. This eliminates the nested ternaries.
I am however keeping the Map usage, as it makes adding new operation modes easier, while not increasing the complexity by that THAT much, mostly due to the low size of the file.

On another note, if i were to return the capital letter Footer, i would be returning the same JSX Element, as the enclosing component, unless you also had a renaming of the component in mind?

.gitignore
andr9528: It is a file generated by Cliptool now, that contains a Record of which files are hidden, with their size and changed values.

src/server/utils/ccgHandlerUtils.ts
andr9528: Good to learn that the Typescript Arrays' `.some(...)` is the quivialant of LINQ's `.Any(...)`.

Have applied the suggest change, with a bit more styling, for readability.

src/client/components/Thumbnail.tsx
andr9528: I completely agree - sent the complaint to Kasper.

Would love to go through all the code and refactor it, to uphold this standard.

src/client/components/Thumbnail.tsx
andr9528: I completely agree - sent the complaint to Kasper.

Would love to go through all the code and refactor it, to uphold this standard.

src/client/components/Thumbnail.tsx
andr9528: I completely agree - sent the complaint to Kasper.

Would love to go through all the code and refactor it, to uphold this standard.

src/client/components/Thumbnail.tsx
andr9528: I completely agree - sent the complaint to Kasper.

Would love to go through all the code and refactor it, to uphold this standard.

src/server/handlers/socketIOServerHandler.ts
andr9528: I am toggleing, wheather the file in question is in the record of hidden files, by either deleting a record entry (turning off) or creating a record entry (turning on).

The method was build together with Anders, where we did extract the responsibility of building the new record out into another function.
Not sure how i'd better split the responsibilities in a more meaningfull way.

src/model/reducers/settingsReducer.ts
andr9528: By returning immediatly AND removing the current single return at the bottom, a default case is needed, which only returns the `nextState`.
If the single return at the bottom remains, the default case is gone. However in this case immediatly returning has the same effect as simly breaking, but at the cost of a longer statement.

If there was just ONE case where the state was not returned, then yes i would be returning in each case. But as it is, it is ALWAYS returned, so having one return statement at the end simplifies the code, as there at the end is only one path out of the function.

src/client/components/Thumbnail.tsx
olzzon: @andr9528 and @LindvedKrvang 
Feel free to rewrite it to your teams guidelines, that has been implemented after this code was written.
But comments as:
"I completely agree - sent the complaint to Kasper."
I'd really be happy without.

src/model/reducers/mediaReducer.ts
andr9528: It is that, or using a type that would fill a couple of lines.
I am also not entirely sure if it would need to be a partial or not.

I would very much like to spend the time putting type of everything, while redoing the code structure.
This would be one of the areas that i'd spent some time improving.

src/client/components/Settings.tsx
```diff
@@ -5,11 +5,15 @@ import { setGenerics } from '../../model/reducers/settingsAction'
...
 import * as IO from '../../model/SocketIoConstants'
 import { TOGGLE_SHOW_SETTINGS } from '../../model/reducers/appNavAction'
+import { OperationMode } from '../../model/reducers/mediaReducer'
 
 // Check if URL has specifiet a channel:
 const channel = new URLSearchParams(window.location.search).get('channel')
 const specificChannel = parseInt(channel) || 0
 
+const OFF_COLOR = { backgroundColor: 'rgb(41, 41, 41)' }
+const ON_COLOR = { backgroundColor: 'rgb(28, 115, 165)' }
```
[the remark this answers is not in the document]
andr9528: Will be extracted to CSS, along with a few other areas with similar created styling code, during my refactor of the code structure.

src/client/components/Settings.tsx
andr9528: Will be gone with the refactor of code structure, as the styling it uses, will be extracted to CSS.

src/client/components/Settings.tsx
```diff
@@ -27,6 +31,23 @@ const selectorColorStyles = {
     singleValue: (styles) => ({ ...styles, color: 'white' }),
 }
 
+function handleToggleEditVisibilityMode() {
```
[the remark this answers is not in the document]
andr9528: If i am in any other mode, it turns on 'Edit Visibility' mode, otherwise i turn off 'Edit Visibility' mode, by setting the mode to the default one, being 'Control'.
Maybe creating a `OperationModeService` class which has function for setting the mode to each posible one, could help with simplifying this.

src/client/components/Thumbnail.tsx
andr9528: @olzzon What i meant with the below comment, was more so, that I was not the one to create the structure, that Lindved, was commenting on. 
>"I completely agree - sent the complaint to Kasper."

As soon as i have checked the remainder of comments on this PR, i will be creating a task in JIRA, wherein the task is to go throught Cliptool, updating the code structure to match how we want it.

src/model/reducers/mediaActions.ts
andr9528: Huh, i missed this one.
Fixed now.

src/client/components/Thumbnail.tsx
andr9528: I completely agree.

Would love to go through all the code and refactor it, to uphold this standard.

src/model/reducers/mediaActions.ts
andr9528: I tried to keep the pattern of function names, from other functions in the file.

A better name might be `builderUpdateHiddenFiles(...)` as the function builds an object, containing the information needed to update hidden files. If that naming scheme is liked, i would go ahead and apply it to a number of function in this file.

src/model/reducers/mediaReducer.ts
andr9528: See my comment on your similar comment.

src/client/css/Footer.css
andr9528: The casing of the CSS classes will be fixed in the restructureing of the code.

As for the general styling, i like the idea, because of less duplication, but i am unsure how to apply it to Cliptool.
How would i go about styling a button on the topbar, and a button under settings? 
Both of the mention buttons has a number of unique styling properties, based on where they are.

Would be very happy if you'd show me.
My ability with CSS is still somewhat limited.

src/client/components/Footer.tsx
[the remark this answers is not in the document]
KvelaGorrrrnio: I don't think that React imports the CSS file automatically. It requires an import statement (either in css or js).

src/client/components/Thumbnail.tsx
```diff
@@ -42,176 +44,196 @@ export const isThumbWithTally = (thumbName): boolean => {
...
             storeUpdate.media[0].output[reduxState.appNav[0].activeTab]
                 ?.mediaFiles
     )
-
-    // Render:
-    if (reduxState.appNav[0].selectView === 0) {
+    const hiddenFiles: Record<string, IHiddenFileInfo> = useSelector(
+        (storeUpdate: any) => storeUpdate.media[0].hiddenFiles
+    )
+    const editVisibilityMode = useSelector(
```
KvelaGorrrrnio, line 55: To indicate that this is a boolean value instead of a callable construct, this could be renamed to `isInEditVisisibilityMode`.

src/client/components/Thumbnail.tsx
```diff
@@ -42,176 +44,196 @@ export const isThumbWithTally = (thumbName): boolean => {
...
+        (storeUpdate: any) =>
+            storeUpdate.media[0].output[reduxState.appNav[0].activeTab]?.operationMode === OperationMode.EDIT_VISIBILITY
+    )
+    const shownFiles: IMediaFile[] = files?.filter(({ name }) => !(name in hiddenFiles))
+    const usedFiles: IMediaFile[] = editVisibilityMode ? files : shownFiles
```
KvelaGorrrrnio, line 60: If we look at the efficiency of this, you are filtering out `hiddenFiles`, even in the case where they are shown. Ideally the `shownFiles` should only be filtered if we are in *Edit Visibility* mode.

src/server/handlers/socketIOServerHandler.ts
KvelaGorrrrnio: You could extract the logic for hiding and showing the files into their own functions like so:
```typescript
type IHiddenFiles = Record<string, IHiddenFileInfo>

function toggleHiddenFile(fileName: string, channelIndex: number, hiddenFiles: IHiddenFiles): IHiddenFiles {
    return isFileHidden(fileName, hiddenFiles)
        ? showFile(fileName, hiddenFiles)
        : hideFile(fileName, channelIndex, hiddenFiles)
}

function isFileHidden(fileName: string, hiddenFiles: IHiddenFiles): boolean {
    return fileName in hiddenFiles
}

function showFile(fileName: string, hiddenFiles: IHiddenFiles): IHiddenFiles {
    const { [fileName], ...newHiddenFiles } = hiddenFiles
   return newHiddenFiles
}

function hideFile(fileName: string, channelIndex: number, hiddenFiles: IHiddenFiles): IHiddenFiles {
    const hiddenFile = getHiddenFileMetadataFromFileName(filename, channelIndex)
    return { ...hiddenFiles, [fileName]: hiddenFile }
}

**Review by KvelaGorrrrnio — COMMENTED** [review]
You have made some good improvements and I found a little more that needs som love ❤️

File: src/client/components/Footer.tsx [hunk 41]
```diff
@@ -0,0 +1,56 @@
...
+    <>
+      <div className='Footer-text'>
+        {footerDescriptions.has(props.operationMode) 
+        ? footerDescriptions.get(props.operationMode).toUpperCase()
+        : 'MISSING DESCRIPTION FOR SELECTED OPERATION MODE!'}
+      </div>        
+      <button onClick={resetOperationMode} className='Footer-button'>
+        {exitButtonText.has(props.operationMode) 
+        ? exitButtonText.get(props.operationMode) 
+        : 'MISSING BUTTON TEXT!'}
```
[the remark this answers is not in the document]
andr9528: Changes Applied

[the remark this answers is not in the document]
andr9528: The map is gone, with the breakdown of the footer into smaller components, like suggested with <Insert OperationMode>Footer.

src/client/components/Thumbnail.tsx
```diff
@@ -5,22 +5,24 @@ import { reduxState } from '../../model/reducers/store'
...
 import { socket } from '../util/SocketClientHandlers'
-import { PGM_LOAD, PGM_PLAY } from '../../model/SocketIoConstants'
+import { PGM_LOAD, PGM_PLAY, TOGGLE_THUMBNAIL_VISIBILITY } from '../../model/SocketIoConstants'
 import { useSelector } from 'react-redux'
 
+interface ThumbnailProps {
+    file: IMediaFile
+}
+
 export const findThumbPix = (fileName: string, channelIndex: number) => {
```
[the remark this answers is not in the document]
andr9528: Added return type and updated to be a function definition instead of constant.

src/client/components/Thumbnail.tsx
```diff
@@ -42,176 +44,195 @@ export const isThumbWithTally = (thumbName): boolean => {
 
 export const Thumbnail = () => {
     // Redux hook:
-    useSelector(
+    const files: IMediaFile[] = useSelector(
         (storeUpdate: any) =>
             storeUpdate.media[0].output[reduxState.appNav[0].activeTab]
                 ?.mediaFiles
     )
```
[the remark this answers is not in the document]
andr9528: We dislike undefined checks, so i will be adding nullish coalescing to the different `useSelector(...)`.

src/client/components/Thumbnail.tsx
andr9528: Renamed to the mentioned name.

src/client/components/Thumbnail.tsx
```diff
@@ -42,176 +44,195 @@ export const isThumbWithTally = (thumbName): boolean => {
...
-                    <div className="boxComponentText" key={index}>
-                        <RenderThumbText item={item} />
-                    </div>
-                ))}
-            </div>
-        )
+        )    
+}
+
+const handleClickMedia = (fileName: string) => {
```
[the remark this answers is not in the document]
andr9528: Added return type and updated to be a function definition instead of constant.

src/client/components/Thumbnail.tsx
```diff
@@ -42,176 +44,195 @@ export const isThumbWithTally = (thumbName): boolean => {
...
+const RenderThumbText = (props: ThumbnailProps) => {
     // Redux hook:
     useSelector(
         (storeUpdate: any) =>
             storeUpdate.media[0].output[reduxState.appNav[0].activeTab]
                 .tallyFile
     )
+    const classNames = [
+        "thumbnail-text-view",
+        isThumbWithTally(props.file.name) ? 'selected-thump' : ''
```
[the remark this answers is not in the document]
andr9528: Fixed

src/client/css/Thumbnail.css
```diff
@@ -32,7 +32,7 @@
 }
 
 .thumbnailImage {
-    border-width: 2px;
+    border-width: 0px;
```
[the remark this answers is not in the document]
andr9528: I must have changed the border to a fixed width on the wrong branch. 
Have changed it so the border always is there, and it is only the color that changes on select - not the border width.

src/server/handlers/socketIOServerHandler.ts
andr9528: Reworked the method to use most of Anders' suggestion.

src/server/handlers/socketIOServerHandler.ts
```diff
@@ -103,6 +152,45 @@ export function socketIoHandlers(socket: any) {
...
+    return hiddenFiles
+}
+
+function buildHiddenFileMetadataFromFileName(
+    fileName: string,
+    channelIndex: number
+): IHiddenFileInfo {
+    const file = reduxState.media[0].output[channelIndex].mediaFiles.find(
+        (file) => file.name.toUpperCase() === fileName.toUpperCase()
+    )
```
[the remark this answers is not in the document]
andr9528: Extracted the piece of code that finds the file to its own function.

On an unrelated note, my VS Code extention says this comment chain is on the function definition (`function buildHiddenFileMetadataFromFileName`...), not on the line accessing the `reduxState`, like it is on GitHub. Weird.

src/server/handlers/socketIOServerHandler.ts
```diff
@@ -17,23 +17,61 @@ import {
...
+                    socketServer.emit(
+                        IO.HIDDEN_FILES_UPDATE,
+                        channelIndex,
+                        updatedHiddenFiles
+                    )
+                } catch {
+                    console.error(
+                        'Error thrown during "TOGGLE_THUMBNAIL_VISIBILITY"'
+                    )
+                }
```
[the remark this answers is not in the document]
andr9528: Re-added the log statement - now using logger! :)
Also includes the error as data in the log statement.

src/client/components/Settings.tsx
```diff
@@ -150,10 +169,17 @@ export const SettingsPage = () => {
...
+                    onClick={handleSettingsPage}
                 >
                     EXIT
                 </button>
+                <button
+                    className="save-button"
+                    onClick={handleToggleEditVisibilityMode}
+                    style={editVisibilityStyle()}
+                >
+                    HIDING
```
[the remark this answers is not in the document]
andr9528: Was a remnant of the topbar having limited space. 
Changed to say 'Edit visibility', in all caps, like the other buttons.

src/client/components/Header.tsx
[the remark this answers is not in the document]
andr9528: Yep i know that the arrow is needed, if arguments has to be passed in. Also think it might need an arrow function, if the return type isn't `void` or `Promise<void>`, but don't quote me on that.

src/client/components/Footer.tsx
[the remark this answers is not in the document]
andr9528: Footer has been split into multiple smaller components. Among those a component whos sole purpose is to contain the logic and constants specific for `OperationMode`.EDIT_VISIBILITY`. This eliminates the use of the map.

src/client/components/Footer.tsx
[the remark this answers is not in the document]
andr9528: With my comment on one of the other footer chains (https://github.com/tv2/casparcg-cliptool/pull/78#discussion_r1116870992), my last comment has become outdated.

src/client/css/Footer.css
andr9528: I very much like that suggestion Anders.

Unsure if it is within the scope, when i am going through and splitting each file, with multiple components, into max one component per file, during UT-210.

src/client/components/Thumbnail.tsx
```diff
@@ -42,176 +44,195 @@ export const isThumbWithTally = (thumbName): boolean => {
...
                 ?.mediaFiles
     )
-
-    // Render:
-    if (reduxState.appNav[0].selectView === 0) {
+    const hiddenFiles: Record<string, IHiddenFileInfo> = useSelector(
+        (storeUpdate: any) => 
+            storeUpdate.media[0].output[reduxState.appNav[0].activeTab]
+                ?.hiddenFiles
+    )
```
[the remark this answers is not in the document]
andr9528: Have extracted it to a function, and applied it to around a dozen lines.

Might be an usefull function to extract further out, than within just Thumbnail.tsx, as other places also does the similar long lines, to get the active output.

Not sure where to place it though.
Maybe a ReduxService class with a handfull of similar helpful functions to gets specific parts from redux, could be usefull. Again, not sure where that should be located, if we want to do something like that.

src/model/reducers/mediaReducer.ts
```diff
@@ -1,22 +1,29 @@
 import * as IO from './mediaActions'
 
-export interface IMediaFile {
+export enum OperationMode {
+    CONTROL = 'control',
+    EDIT_VISIBILITY = 'edit_visibility',
+}
```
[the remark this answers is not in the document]
andr9528: The `OperationMode` is used as a type on `IOutput`, from where it is then used multiple places. Not sure where a more fitting places would be.

src/model/reducers/settingsReducer.ts
andr9528: I think it would be nice to get input from @olzzon and @RasmusAlbrektsen on this one.
If either, also prefer the 'return-immediately' in this scenario, then i will change it back.

andr9528: After a chat with Kasper, i found that the original structure that he had developed, was a replica of the structure used in Redux documentation, thus it now makes more sense to be why it was used. I will be reverting this switch change, anywhere where Redux is part of it.

KvelaGorrrrnio (approved):
I have approved on the assumption that the following comments are addressed.
Nice work! 🚀

src/client/components/Settings.tsx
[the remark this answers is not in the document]
LindvedKrvang: No matter what mode you are in you are just emitting an event here.
What the listeners of said event do (which may very well be toggling) has nothing to do with the`handleToggleEditVisibilityMode()` function here, hence the name of this function should not indicate anything else than emitting an event.

src/client/components/Thumbnail.tsx
LindvedKrvang: > sent the complaint to Kasper.

Doesn't matter who wrote it first

src/client/components/Thumbnail.tsx
LindvedKrvang: Then do it - it's a quick boyscout

src/client/components/Thumbnail.tsx
LindvedKrvang: Then do it - it's a quick boyscout

src/client/components/Thumbnail.tsx
LindvedKrvang: Then do it - it's a quick boyscout

src/client/components/Thumbnail.tsx
LindvedKrvang: Then do it - it's a quick boyscout

src/model/reducers/mediaActions.ts
LindvedKrvang: > A better name might be buildUpdateHiddenFiles(...)

No building is happening. It's simply just retuning a new object

src/model/reducers/settingsReducer.ts
LindvedKrvang: > If there was just ONE case where the state was not returned, then yes i would be returning in each case. But as it is, it is ALWAYS returned, so having one return statement at the end simplifies the code, as there at the end is only one path out of the function.

I disagree. As Anders points out, by not returning you indicate that more is going to happen so you have to read the rest of the method. It doesn't get much simpler than a return statement saying "now we are done"

src/model/reducers/mediaReducer.ts
LindvedKrvang: Read your comment. You should keep the return statements

src/server/utils/ccgHandlerUtils.ts
LindvedKrvang: FYI, most of the operations of LINQ is found on Array

src/server/handlers/socketIOServerHandler.ts
```diff
@@ -102,6 +148,65 @@ export function socketIoHandlers(socket: any) {
...
+    return buildHiddenFileMetadata(file)
+}
+
+function findFile(fileName: string, channelIndex: number): IMediaFile {
+    return reduxState.media[0].output[channelIndex].mediaFiles.find(
+        (file) => file.name.toUpperCase() === fileName.toUpperCase()
+    )
+}
+
+function buildHiddenFileMetadata(file: IMediaFile): IHiddenFileInfo {
```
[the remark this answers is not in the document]
LindvedKrvang: Nothing is being built here. At *most* your are wrapping the parameter in an object.
I would suggest to use `getMetaData(file: IMediaFile)` which clearly indicates that you are extracting the meta data from the file your are passing.

Also, no reason to include 'Hidden' in the name. The function doesn't care about if the file is hidden or not, it just need to be a file.

.gitignore
LindvedKrvang: Why not just use localStorage?

src/client/components/Footer/EditVisibilityFooter.tsx
```diff
@@ -0,0 +1,18 @@
+import React from "react"
+import { FooterContent } from "./FooterContent"
+
+const buttonText = 'Exit Edit Visibility'
+const descriptionText = 'You are currently editing the visibility of files.'.toUpperCase()
```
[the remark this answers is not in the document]
andr9528: Have been renamed to SNAKE_CASE.

src/client/components/Thumbnail.tsx
```diff
@@ -40,178 +49,183 @@ export const isThumbWithTally = (thumbName): boolean => {
...
-        <div>
-            <RenderThumbPix item={props.item} />
 
+    const hiddenFiles: Record<string, IHiddenFileInfo> = useSelector(
+        (storeUpdate: any) => 
+            storeUpdate.media[0].hiddenFiles
+    ) ?? {}
+
+    const classNames = [
+        "thumb",
```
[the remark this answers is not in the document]
andr9528: Think those were copied directly from the assignment to `className`, where it does make use of double quotes. Have updated it in 3 locations to use single quotes.

src/model/reducers/mediaReducer.ts
andr9528: Have added what you suggested Anders, with a slight improvement, from an `unknown[]` to an `IOutput[]`.

src/server/handlers/socketIOServerHandler.ts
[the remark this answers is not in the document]
andr9528: Have renamed the commented on method to `getMetaData(...)`.

src/model/reducers/mediaReducer.ts
LindvedKrvang: > It is that, or using a type that would fill a couple of lines.

It's Typescript so just write the types needed for your actual function and nothing more (as Anders points out). 
Ideally there would be an interface that all States uses and the same for Actions

src/model/reducers/mediaReducer.ts
```diff
@@ -1,26 +1,34 @@
...
-    changed: number
-    size: number
-    thumbnail?: any
+export interface IThumbFile extends ICommonFile {
+    thumbnail?: string
 }
 export interface IMedia {
     output: IOutput[]
     folderList: string[]
+    hiddenFiles: Record<string, IHiddenFileInfo>
```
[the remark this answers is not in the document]
andr9528: I tried with a Map at first, but couldn't get it to work, so Anders came and helped me, and together we found out that Redux doesn't support maps. Redux does however support Records.

src/model/reducers/mediaReducer.ts
```diff
@@ -1,26 +1,34 @@
...
 }
 
-export interface IThumbFile {
-    name: string
-    type: string
-    changed: number
-    size: number
-    thumbnail?: any
+export interface IThumbFile extends ICommonFile {
+    thumbnail?: string
```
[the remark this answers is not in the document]
andr9528: Not sure why Kasper originally made that attribute optional. Removing the ability for it to be optional doesn't appear to break anything.

As for moving it to the `ICommonFile` interface, i have  some questions that would need answer before doing that. In case it is moved, then...
- I take that `thumbnailList` on `IOutput` should have its type changed to `ICommonFile`?

When media files are retrived from CCG, they do not include the thumbnails, those are retrieved in a seperate call to CCG. The code would have to be updated to merge the thumbnail data from the one call, into the media files from the other call.

This is well out of scope for this task, i'd say, even if considered as boy-scouting.

src/model/reducers/mediaReducer.ts
```diff
@@ -1,26 +1,34 @@
...
     duration: number
 }
 
-export interface IThumbFile {
-    name: string
-    type: string
-    changed: number
-    size: number
-    thumbnail?: any
+export interface IThumbFile extends ICommonFile {
```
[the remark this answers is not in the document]
andr9528: Has been renamed.

src/model/reducers/mediaReducer.ts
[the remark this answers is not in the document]
olzzon: @andr9528 I can't recall why i made it optional. 
Could be related to whether the CasparCG scanner is running or not, or som other case. But it could also be an old reminiscence of something.
As you point out the reason for having it separate is that it's received separate from CCG. The other plan for it, was as far as i remember to have all media available on the server, but only the one needed in the client. But it's been a long time since i looked at it, so if you wan't to know for sure, I need to spend a couple of hours digging in to it.

src/model/reducers/mediaReducer.ts
```diff
@@ -1,26 +1,34 @@
...
+    CONTROL = 'control',
+    EDIT_VISIBILITY = 'edit_visibility',
+}
+
+export interface IHiddenFileInfo {
+    changed: number
+    size: number
+}
+
+interface ICommonFile extends IHiddenFileInfo {
```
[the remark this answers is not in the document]
andr9528: Have updated the name to `IFile`. Cleaning up the use of 'I' on interfaces will be handled during UT-210.

Attributes on `HiddenFileInfo` are data relevant to a hidden file, and nothings more. The Attributes of `HiddenFileInfo` also need to exist on the different files (`IMediaFile / IThumbnailFile`), as it is these attributes that gets extracted and persisted to figure out which version of a file is saved.

Attributes on `IFile` are data relevant to `IMediaFile / IThumbnailFile`, while in Cliptool. These attributes on `IFile` are irrelevant to determin the version of the file, when checking if the file has changed outside Cliptool.

src/model/reducers/mediaReducer.ts
```diff
@@ -1,26 +1,34 @@
 import * as IO from './mediaActions'
 
-export interface IMediaFile {
+export enum OperationMode {
+    CONTROL = 'control',
+    EDIT_VISIBILITY = 'edit_visibility',
+}
+
+export interface IHiddenFileInfo {
```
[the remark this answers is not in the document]
andr9528: At first i considered and tried to do it with a attribute as you suggest.
But then i reached the point where i needed to persist that a specific file is hidden, between restart of Cliptool.
If the attribute had stayed there, then i'd need to persist the whole file information, which is more than is needed to figure out which file is hidden. 

Files exist as an array on each output, where each output can have the same files as another, but as a copy.
The fifth acceptance criteria also meant that keeping it as an attribute on the file, would mean that i'd have to find all outputs where that file existed and update the attribute on those too.
Because of this, after a chat with Anders, we came up with a single record saved on `IMedia`, thus working accros multiple outputs.

src/model/reducers/mediaReducer.ts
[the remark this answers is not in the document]
andr9528: @olzzon I'm not going to be moving that attribute for this PR. 
It would be a nice-to-know for in the future, but no rush.

Also that was a quick response from you. I didn't even ping you.

src/server/handlers/CasparCgHandler.ts
```diff
@@ -274,6 +289,31 @@ const outputExtractFiles = (allFiles: any, outputIndex: number) => {
     }
 }
 
+function checkHiddenFilesChanged(files: IMediaFile[]) {
+    let needsUpdating = false
+    const hiddenFiles = reduxState.media[0].hiddenFiles
+    for (const key in hiddenFiles) {
+        const hidden = hiddenFiles[key]
```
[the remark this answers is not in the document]
andr9528: Updated name to `hiddenFileInfo`, and added types to this line and the one 2 lines above.

src/client/css/App.css
```diff
@@ -1,13 +1,27 @@
 /* Main CSS file */
+html {
+    height: 100%;
```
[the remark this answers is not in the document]
andr9528: Removed a few that seemed to be unneded. The remaining is needed to fill out the background, otherwise an ugly white bar is at the buttom of the page.

src/client/components/Thumbnail.tsx
```diff
@@ -5,26 +5,35 @@ import { reduxState } from '../../model/reducers/store'
...
-export const isThumbWithTally = (thumbName): boolean => {
+// TODO: during UT-210, figure out the correct type to apply instead of 'any'.
+function getActiveOutput(store: any, channelIndex: number = -1): IOutput {
+    const activeTab: number = channelIndex === -1 
+    ? reduxState.appNav[0].activeTab 
+    : channelIndex
+    return store.media[0].output[activeTab]
+}
+
+export function findThumbPix(fileName: string, channelIndex: number): string {
```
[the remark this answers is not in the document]
andr9528: Renamed to `findThumbnail`.

src/client/components/Thumbnail.tsx
```diff
@@ -5,26 +5,35 @@ import { reduxState } from '../../model/reducers/store'
...
+    const thumb = getActiveOutput(reduxState, channelIndex)?.thumbnailList
+        .find(
+            (item: IThumbFile) => item.name.toUpperCase() === (fileName.toUpperCase())
+        )
+    return thumb?.thumbnail ?? ''
+}
+
+export const isThumbWithTally = (thumbName: string): boolean => {
```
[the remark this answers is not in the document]
andr9528: Not sure why Kasper added export to `isThumbWithTally`. If he had only one component per file, he'd need it to be exported as it is used by multiple components in this file. As it is, it isn't needed, and Cliptool builds just fine, if it is not exported.

In the case of `findThumbPix` (now renamed to `findThumbnail`), it is also used by the header. Havn't looked deeply into where, what or why. When splitting the file into multiple files for UT-210, this will be fixed.

src/client/components/Footer/FooterContent.tsx
```diff
@@ -0,0 +1,22 @@
...
+interface FooterContentProps {
+  descriptionText: string
+  buttonText: string
+  onClick: () => void
+}
+
+export function FooterContent(props: FooterContentProps): JSX.Element {
+  return (
+  <div className='Footer-flex'>
+    <div className='Footer-text'>
```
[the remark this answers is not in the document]
andr9528: The file in question (and lines) have had mayor changes performed on the branch for UT-211. 

Needs to question Kasper/Jacob/Kevin if we want to release the new and improved tabs too. If not, then changes to the Footer should very likely be cherry-picked onto this branch.

src/client/components/Thumbnail.tsx
```diff
@@ -17,22 +17,32 @@ interface ThumbnailProps {
...
         )
     return thumb?.thumbnail ?? ''
 }
 
-export const isThumbWithTally = (thumbName: string): boolean => {
-    // convert to uppercase, handle windows double slash + backslash and remove file extension:
-    const tallyFileName = getActiveOutput(reduxState)
+function isThumbnailWithTallyOnAnyOutput(thumbnailName: string): boolean {
+    const outputs = reduxState.media[0].output
+    for (const output of outputs) {
```
LindvedKrvang, line 35: In general you should try to avoid the use of loops (they do have their place, but it's becoming more rare). You can use `Array.some()` for the same behaviour
```
function isThumbnailWithAllayOnAnyOutput(...): boolean {
  return reduxState.media[0].ouput.some(output => getCleanTallyFile(output) === thumbnailName)
}
```

src/client/components/Thumbnail.tsx
```diff
@@ -91,8 +105,11 @@ function handleClickMedia(fileName: string): void {
 }
 
 function toggleVisibility(fileName: string) {
-    if (isThumbWithTally(fileName))
+    if (isThumbnailWithTallyOnAnyOutput(fileName)) {
+        alert('Unable to hide, as file is in use somewhere.')
```
LindvedKrvang, line 109: I'm taking a guess here and assume you want it to say `as a file`?

src/client/css/App.css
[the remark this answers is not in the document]
LindvedKrvang: That still worries me and suggest to me that there is somewhere else the CSS isn't right, since this is needed...

src/model/reducers/mediaReducer.ts
```diff
@@ -5,35 +5,35 @@ export enum OperationMode {
...
 }
 
-export interface IHiddenFileInfo {
+export interface HiddenFileInfo {
     changed: number
     size: number
 }
 
-interface ICommonFile extends IHiddenFileInfo {
+interface IFile extends HiddenFileInfo {
```
LindvedKrvang, line 13: Since you are already here refactoring, you should also remove the "I" prefix

src/client/components/Settings.tsx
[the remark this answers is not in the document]
andr9528: Renamed to `emitSetOperationMode`. Still not entirely sure if that name is descriptive enough.

src/client/components/Thumbnail.tsx
andr9528: The task UT-210 was created to go through and extract these functional components to their own files, along with some over structural things. It isn't only in this file where this issue persists, although this is the main offender.

src/client/components/Thumbnail.tsx
andr9528: The alert is triggered when a user clicks to try and hide a file which currently is in use on another output. Thus it is 'Unable to hide, as **the** file is in use somewhere.' Added a 'the' to further specify that the file in question is the one clicked.

src/client/components/Thumbnail.tsx
andr9528: Have applied the suggested change.

In the case of the loop, it did exit early, just as the `.some(...)` does, the loop just had a larger footprint.

src/model/reducers/mediaReducer.ts
andr9528: 'I' prefix has been removed.

LindvedKrvang (approved):

PR #80, opened 2023-02-28
chore: Set bundled NODE_ENV default to production.

KvelaGorrrrnio:
When ClipTool is run from a packaged executable, the NODE_ENV should be set to `production` if not already specified through environment variables.


andr9528:
I have packaged and testet locally that changes in this PR appears to be working as intended.

I ran Cliptool both as a standalone .exe and through CasperCG Launcher.
In both cases the logs i saw was using the (json) structure applied while the NODE_ENV is 'production'.

andr9528 (approved):
Looking forward to getting these changes in. The error related to capital start letter of 'hiddenFileStorage.ts' is annoying me each time i switch branch...

package.json
```diff
@@ -27,9 +27,9 @@
         "build-server": "tsc -p ./src/server",
         "build-gateway": "tsc -p ./src/gateway",
         "build-client": "cross-env NODE_ENV=production TS_NODE_PROJECT=\"./src/client/tsconfig.json\" webpack && copyfiles -f ./src/client/index.html ./build/client/",
-        "start": "node ./build/server/index.js",
-        "start-dev": "node --inspect ./build/server/index.js",
-        "start-local": "cross-env LOG_LEVEL=trace node --inspect ./build/server/index.js",
+        "start": "cross-env NODE_ENV=production node ./build/server/index.js",
```
LindvedKrvang, line 30: So if I run a `yarn start` on my local machine, I would run it as production?

LindvedKrvang (approved):

package.json
andr9528: Yes, atleast for the sake of logging, from my understanding. Not sure what else is effected by the `NODE_ENV=production`, if anything.

RasmusAlbrektsen (approved):

PR #83, opened 2023-04-19
UT-215: lowers brightness of tab color by using same color as active buttons.

andr9528:
Note that this PR points towards the refactor branch.

2 of the acceptance criteria for this task has been solved during UT-210, as the areas related to them had other issues that was solved, and hit them. The hit ones are the following:
- Size of thumbnail in textview is equal to the size in the older version
- Border colour only on selected file

This PR solves the last acceptance criterion.


src/client/components/settings/settings-actions/settings-actions.tsx
```diff
@@ -19,7 +19,7 @@ export default function SettingsActions(): JSX.Element {
   const operationMode = useSelector(
     (state: State) => settingsService.getOutputSettings(state.settings, activeTab)?.operationMode)
   
-  const buttonCss = "save-button"
+  const buttonCss = "c-settings-actions"
```
KvelaGorrrrnio, line 22: The plural form seems incorrect, since it is used on the buttons.
I also wonder how much the styling on this differs from the "default" button. It could possibly be removed.

src/client/css/index.scss
```diff
@@ -14,6 +14,7 @@ body {
 
     --cued-color: green;
     --selected-color: red;
+    --active-color: rgb(28, 115, 165);
```
KvelaGorrrrnio, line 17: Is the color dimmed enough?

src/client/components/settings/settings-actions/settings-actions.tsx
andr9528: have renamed the CSS class to "settings-action". Note that it doesn't have the "c-" any longer as it doesn't match the name of the component that uses it anylonger. Don't know if that is the correct way.

As for how much of the styling differs from default button, i'd say alot. It has 10 different lines, where only the font type doesn't appear to have much effect being there or not. 
Perhabs some of the styling could/should be moved and applied to button dirrectly, much like has been done to input.

KvelaGorrrrnio: As you talked about last week, it probably will make sense to have a separate style refactor PR. So It is fine for now :)

KvelaGorrrrnio (approved):

RasmusAlbrektsen (approved):

LindvedKrvang (approved):

PR #86, opened 2023-05-26
UT-214: Clear file/folder paths targeting invalid paths

andr9528:
I'd highly suggest to read the comments on the task as well, for some more insigt into the process/reasoning being the changes.

Notice that the base of this, is the refactor branch, so merging can only happen after the Refactor PR (#81) has been merged.


src/model/services/settings-service.ts
```diff
@@ -43,6 +43,38 @@ class SettingsService {
...
+    ): OutputSettings {
+        const outputSettings = { ...originalOutputSettings }
+        if (
+            outputSettings.selectedFileName === '' &&
+            outputSettings.cuedFileName === '' &&
+            outputSettings.folder === ''
+        ) {
+            return outputSettings
+        }
+        const pathExist = allFiles.find((file) => {
```
KvelaGorrrrnio, line 59: `pathExist` conveys that this is a boolean, but it is `MediaFile | undefined`.

src/server/handlers/caspar-cg-handler.ts
```diff
@@ -265,24 +267,31 @@ function startTimeEmitInterval() {
 }
 
 async function startFileChangesPollingInterval(): Promise<void> {
-    await pollFileChanges()
+    await pollThumbnailChanges()
```
KvelaGorrrrnio, line 270: Why shouldn't we pollFileChanges?

src/server/handlers/caspar-cg-handler.ts
```diff
@@ -265,24 +267,31 @@ function startTimeEmitInterval() {
...
     setInterval(() => {
         if (!waitingForCcgResponse) {
             waitingForCcgResponse = true
             pollFileChanges().then(() => {
-                waitingForCcgResponse = false
+                pollThumbnailChanges().then(() => {
```
KvelaGorrrrnio, line 275: Are these required to be executed sequentially?

src/model/services/settings-service.ts
```diff
@@ -43,6 +43,38 @@ class SettingsService {
         return settingsState.generics.outputSettings[channelIndex]
     }
 
+    public fixInvalidUsedPaths(
```
KvelaGorrrrnio, line 46: What is this function trying to solve?

src/server/handlers/caspar-cg-handler.ts
```diff
@@ -356,7 +361,50 @@ async function loadFileList(): Promise<void> {
         })
 }
 
-function outputExtractFiles(allFiles: MediaFile[], outputIndex: number): void {
+function extractFilesToOutputs(allFiles: MediaFile[]): void {
+    const outputs = mediaService.getOutputs(state.media)
+    if (outputs.length !== state.settings.ccgConfig.channels.length) {
+        logger.warn(
+            `Expected '${state.settings.ccgConfig.channels.length}' Outputs but had '${outputs.length}'`
```
KvelaGorrrrnio, line 368: Why are the numbers in single-quotes?

src/server/handlers/caspar-cg-handler.ts
```diff
@@ -356,7 +361,50 @@ async function loadFileList(): Promise<void> {
         })
 }
 
-function outputExtractFiles(allFiles: MediaFile[], outputIndex: number): void {
+function extractFilesToOutputs(allFiles: MediaFile[]): void {
```
KvelaGorrrrnio, line 364: It does not look like this is being extracted, but rather that the _media_ files are merged into the outputs.

src/server/handlers/caspar-cg-handler.ts
```diff
@@ -416,29 +461,32 @@ async function loadThumbnailImage(
...
+            const outputMedia = thumbnails.filter(
+                (thumbnail: ThumbnailFile) => {
+                    return isFolderNameEqual(
+                        thumbnail?.name,
+                        settingsService.getOutputSettings(
+                            state.settings,
+                            channelIndex
+                        ).folder
+                    )
+                }
```
KvelaGorrrrnio, line 476: There are a number of syntactical and performance optimizations that can be applied here. If you extract the output folder before doing the filter, there will be a lot less lookups for the output folder. If you apply that along with using shorthand notation it becomes:
```typescript
            const outputFolder = settingsService.getOutputSettings(state.settings, channelIndex).folder
            const outputMedia = thumbnails.filter(thumbnail => isFolderNameEqual(thumbnail.name, outputFolder))
```

src/server/handlers/caspar-cg-handler.ts
```diff
@@ -356,7 +361,50 @@ async function loadFileList(): Promise<void> {
...
+            `Expected '${state.settings.ccgConfig.channels.length}' Outputs but had '${outputs.length}'`
+        )
+    }
+    outputs.forEach((output: Output, outputIndex: number) => {
+        outputExtractFiles(allFiles, outputIndex, output)
+    })
+}
+
+function fixInvalidUsedPathsInSettings(allFiles: MediaFile[]): void {
+    const fixedPaths = settingsService
```
KvelaGorrrrnio, line 377: What does it mean to fix? And what is a fixed path?

KvelaGorrrrnio (commented):
I am not entirely sure what this fixes – so please enlighten me :smile:

andr9528:
> I am not entirely sure what this fixes – so please enlighten me 😄

@KvelaGorrrrnio If Cliptool has selected / cued a file for playing, that is then deleted regardless of if Cliptool is running or not, the targeted file will remain in the settings file, even though it is technically an invalid target now.
This PR fixes this by checking if the targeted file/folders exist when Cliptool boots. If the file/folder doesn't exist Cliptool remvoves the invalid reference in the settings file.

~Now that i think of it, the code that does that cleanup, should propably also be run if the media files change, to clear the settings file of the invalid path, should a file be deleted that Cliptool has set to run / cue while running.~ Never mind this part, as the code already should do it, with how i implemented it.

Might not be critical to clear the invalid target at the moment, but i could see it being an issue, if not fixed before UT-234, as CasparCG might be asked to play a non-existant file.

Hope that made sense, and helped.

src/server/handlers/caspar-cg-handler.ts
andr9528: Not entirely sure, but i don't think so.
Have changed the function given to the interval to be async, and used await instead,

Also then moved setting `waitingForCcgResponse` to false to just after the awaits.
That should afaik ensure it gets set to false, even if the awaited ones fails somehow.

src/server/handlers/caspar-cg-handler.ts
andr9528: Hmm, i think i had some reasoning for it, but can't fully remember what. 
Have re-added polling for file changes, followed by the existing code.

src/server/handlers/caspar-cg-handler.ts
```diff
@@ -265,24 +267,31 @@ function startTimeEmitInterval() {
...
-    await pollFileChanges()
+    await pollThumbnailChanges()
     setInterval(() => {
         if (!waitingForCcgResponse) {
             waitingForCcgResponse = true
             pollFileChanges().then(() => {
-                waitingForCcgResponse = false
+                pollThumbnailChanges().then(() => {
+                    waitingForCcgResponse = false
+                })
```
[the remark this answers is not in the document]
andr9528: Afaik the changes made for the follow comment, should fix this.
https://github.com/tv2/casparcg-cliptool/pull/86/files#r1252793629

src/server/handlers/caspar-cg-handler.ts
andr9528: A bad habbit of mine. Gone now.

src/server/handlers/caspar-cg-handler.ts
andr9528: Have applied the suggested changes here, and in one other similar place.

src/model/services/settings-service.ts
andr9528: Renamed to `mediaFile`.

src/model/services/settings-service.ts
andr9528: It clears `cuedFileName`, `selectedFileName` and `folder` on an `outputSettings`, if they target a file/folder that doesn't exist any longer, i.e it clears invalid targeted paths, which it has now been renamed to better describe (`clearInvalidTargetedPaths(...)`).

src/server/handlers/caspar-cg-handler.ts
andr9528: Hmm, it extractes the files from the folder matching the one the output is configred for, and assign them to the specific output.

If it didn't do the filtering, then i'd agree that it just merges, but that is not the case.

src/server/handlers/caspar-cg-handler.ts
andr9528: An `outputSettings` with `fixedPaths` no longer points to files/folders that have been deleted.

 Have renamed the constant to `outputSettingsWithfixedPaths`, not sure if that helps, or is a sign that something needs to be reworked.

src/server/handlers/caspar-cg-handler.ts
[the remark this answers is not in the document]
KvelaGorrrrnio: The change you reference does not fix the issue. You still have a sequence of statements, where the sequence will prematurely abort if one step fails.
Since `waitingForCcgResponse` is the last statement, it will not be executed in case of failure from either of the previous statements. 
And in the case of failure, the rejected promise will not be caught. That is still something

The right approach would be to use finally syntax, either via Promise handling or try-catch-finally handling. It will solve both issues.

src/server/handlers/caspar-cg-handler.ts
```diff
@@ -370,22 +372,33 @@ async function loadFileList(): Promise<void> {
...
         )
+    }
+    outputs.forEach((output: Output, outputIndex: number) => {
+        outputExtractFiles(allFiles, outputIndex, output)
     })
-    const mediaFiles = new ReduxMediaService().getOutput(
-        state.media,
+}
+
+function outputExtractFiles(
```
KvelaGorrrrnio, line 387: The function name seems a bit odd. Are we outputting the function extractFiles? And what does output mean in this context? Is it printing/logging or emitting some data or something third?

src/server/handlers/caspar-cg-handler.ts
```diff
@@ -397,6 +410,33 @@ function outputExtractFiles(allFiles: MediaFile[], outputIndex: number): void {
     }
 }
 
+function fixInvalidUsedPathsInSettings(allFiles: MediaFile[]): void {
+    const outputSettingsWithfixedPaths = new ReduxSettingsService()
```
KvelaGorrrrnio, line 414: `fixed` should be uppercased.

src/server/handlers/caspar-cg-handler.ts
```diff
@@ -397,6 +410,33 @@ function outputExtractFiles(allFiles: MediaFile[], outputIndex: number): void {
...
+    ) {
+        logger.warn(
+            'Removing some invalid paths from settings, that likely exist due to folders/files being deleted while off.'
+        )
+        const genericSettings = { ...state.settings.generics }
+        genericSettings.outputSettings = outputSettingsWithfixedPaths
+        reduxStore.dispatch(setGenerics(genericSettings))
+        assignThumbnailsToOutputs()
+        new SettingsPersistenceService().save()
+    }
```
KvelaGorrrrnio, line 437: You can invert the condition to make it a guard, which will remove an execution scope.
And if you extract the logic out into their own functions, this function will be easier to read and understand.

src/shared/services/redux-settings-service.ts
```diff
@@ -45,6 +45,38 @@ export class ReduxSettingsService {
...
+        })
+        if (!mediaFile) {
+            outputSettings.cuedFileName = ''
+            outputSettings.selectedFileName = ''
+        }
+        if (!mediaState.folders.includes(outputSettings.folder)) {
+            outputSettings.folder = ''
+        }
+
+        return outputSettings
```
KvelaGorrrrnio, line 73: Extracting the logic and relevant conditions out into appropriately named methods, will make the flow of this function more digestable.

src/server/handlers/caspar-cg-handler.ts
```diff
@@ -280,23 +280,29 @@ function startTimeEmitInterval() {
 
 async function startFileChangesPollingInterval(): Promise<void> {
     await pollFileChanges()
-    setInterval(() => {
+    await pollThumbnailChanges()
+    setInterval(async () => {
```
KvelaGorrrrnio, line 284: Just a note: Currently we setup an interval here, but the handle created is not stored, so we have no way of clearing the interval. And if we call this function multiple times, we will get a lot of unnecessary work to do.

src/server/handlers/caspar-cg-handler.ts
```diff
@@ -280,23 +280,29 @@ function startTimeEmitInterval() {
...
+    await loadFileList()
+}
+
+async function pollThumbnailChanges(): Promise<void> {
+    logger.trace('Polling for Thumbnail Changes.')
     const newThumbnails = await loadCcgThumbnails()
     if (newThumbnails.hasChanges) {
         thumbnails = newThumbnails.thumbnails
         assignThumbnailsToOutputs()
     }
```
KvelaGorrrrnio, line 305: Just a small optimization: This can be changed into a guard.
```suggestion
    if (!newThumbnails.hasChanges) {
        return
    }
    thumbnails = newThumbnails.thumbnails
    assignThumbnailsToOutputs()
```

src/server/handlers/caspar-cg-handler.ts
```diff
@@ -280,23 +280,29 @@ function startTimeEmitInterval() {
...
+            await pollFileChanges()
+            await pollThumbnailChanges()
+            waitingForCcgResponse = false
         }
     }, 3000)
 }
 
 async function pollFileChanges(): Promise<void> {
+    logger.trace('Polling for File Changes.')
+    await loadFileList()
```
KvelaGorrrrnio, line 296: Should there be checked for changes in the files loaded, like for thumbnails?

src/server/handlers/caspar-cg-handler.ts
```diff
@@ -280,23 +280,29 @@ function startTimeEmitInterval() {
...
+}
+
+async function pollThumbnailChanges(): Promise<void> {
```
KvelaGorrrrnio, line 299: Just a note: `poll` seems to say the same as `get` in this context. Poll might points toward long-polling, which is not the case here.

src/server/handlers/caspar-cg-handler.ts
KvelaGorrrrnio: One could bundle the promises, and await both with `Promise.all`. But I don't think it is necessary for that optimization. So I'll resolve this conversation. I have commented on the relevant comment regarding `waitingForCcgResponse` begin set to false.

src/server/handlers/caspar-cg-handler.ts
```diff
@@ -441,22 +481,15 @@ async function loadThumbnailImage(
...
-            const outputThumbnailList = new ReduxMediaService().getOutput(
-                state.media,
+        .forEach((output: Output, channelIndex: number) => {
+            const outputFolder = new ReduxSettingsService().getOutputSettings(
+                state.settings,
                 channelIndex
-            ).thumbnailList
+            ).folder
+            const outputMedia = thumbnails.filter((thumbnail: ThumbnailFile) =>
+                isFolderNameEqual(thumbnail?.name, outputFolder)
```
KvelaGorrrrnio, line 490: Why is optional chaining use here? When will a thumbnail not have a name?

src/server/handlers/caspar-cg-handler.ts
```diff
@@ -441,22 +481,15 @@ async function loadThumbnailImage(
...
+            )
+            const outputThumbnailList = output.thumbnailList
```
KvelaGorrrrnio, line 492: Why is this stored in a local variable?

src/server/handlers/caspar-cg-handler.ts
```diff
@@ -370,22 +372,33 @@ async function loadFileList(): Promise<void> {
...
+        state.settings,
         outputIndex
-    ).mediaFiles
+    ).folder
+    const outputMedia = allFiles.filter(
+        (file) =>
+            isFolderNameEqual(file.name, outputFolder) &&
+            !isAlphaFile(file.name)
+    )
+    const mediaFiles = output.mediaFiles
```
KvelaGorrrrnio, line 401: Why is this stored in a local variable?

KvelaGorrrrnio (changes requested):

src/server/handlers/caspar-cg-handler.ts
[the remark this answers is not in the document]
andr9528: Moved the awaits into a try, and the reset of boolean into a finally.

src/server/handlers/caspar-cg-handler.ts
```diff
@@ -280,23 +280,29 @@ function startTimeEmitInterval() {
...
+    await pollThumbnailChanges()
+    setInterval(async () => {
         if (!waitingForCcgResponse) {
             waitingForCcgResponse = true
-            pollFileChanges().then(() => {
-                waitingForCcgResponse = false
-            })
+            await pollFileChanges()
+            await pollThumbnailChanges()
+            waitingForCcgResponse = false
```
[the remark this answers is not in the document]
andr9528: Done.

src/server/handlers/caspar-cg-handler.ts
```diff
@@ -370,22 +372,33 @@ async function loadFileList(): Promise<void> {
...
+    allFiles: MediaFile[],
+    outputIndex: number,
+    output: Output
+): void {
+    const outputFolder = new ReduxSettingsService().getOutputSettings(
+        state.settings,
         outputIndex
-    ).mediaFiles
+    ).folder
+    const outputMedia = allFiles.filter(
```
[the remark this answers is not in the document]
andr9528: `output.mediaFiles` contains what we currently have, and `outputMedia` (now `outputMediaFiles`) contains the new versions of the files.

If they are different, then the new ones overwrite the old ones.

Have removed saving of `output.mediaFiles` to a new constant.

src/server/handlers/caspar-cg-handler.ts
andr9528: Renamed to `extractFilesToOutput`.

src/server/handlers/caspar-cg-handler.ts
andr9528: Fixed.

src/server/handlers/caspar-cg-handler.ts
andr9528: Have inverted it, and extracted the innards of the if to a new function called `saveFixedPathsSettings`.

src/shared/services/redux-settings-service.ts
andr9528: Have done that now.

src/server/handlers/caspar-cg-handler.ts
andr9528: Done

src/server/handlers/caspar-cg-handler.ts
andr9528: Have renamed them to say `get` instead.

src/server/handlers/caspar-cg-handler.ts
andr9528: This timeout should now have logic to ensure that only one instance of the timeout runs.

Unsure if something similar should be done for the timeout in the `startTimeEmitInterval` function.
And if so, how about the subscribtion to the `reduxStore`?

src/server/handlers/caspar-cg-handler.ts
andr9528: It isn't any longer :P

src/server/handlers/caspar-cg-handler.ts
andr9528: I think that was a leftover from before refactor.
Have removed it now, with no ill effect from what i can tell.

src/server/handlers/caspar-cg-handler.ts
andr9528: Afaik, those checks are done further down the call stack.

src/server/handlers/caspar-cg-handler.ts
```diff
@@ -279,30 +280,38 @@ function startTimeEmitInterval() {
...
 async function startFileChangesPollingInterval(): Promise<void> {
-    await pollFileChanges()
-    await pollThumbnailChanges()
-    setInterval(async () => {
-        if (!waitingForCcgResponse) {
-            waitingForCcgResponse = true
-            await pollFileChanges()
-            await pollThumbnailChanges()
-            waitingForCcgResponse = false
+    if (changesTimeout) {
```
KvelaGorrrrnio, line 283: Timeout indicates that it is a single-off effect. So since we are storing an interval in this, it could be named either `fileChangesInterval` or `fileChangesTimer`.

src/shared/services/redux-settings-service.ts
```diff
@@ -51,13 +51,55 @@ export class ReduxSettingsService {
...
+            scaleY: number
+            webUrl: string
+            loopState: boolean
+            mixState: boolean
+            manualStartState: boolean
+            webState: boolean
+            operationMode: import('c:/Code/Tv2/Sofie Related/casparcg-cliptool/src/shared/models/settings-models').OperationMode
+            selectedFileName: string
+            cuedFileName: string
+        }
```
KvelaGorrrrnio, line 78: There should not be paths referencing the specifics of your local environment.

Don't we have a type for outputSettings? In the unlikely case that we don't have a type for it, then this function is only dependent on the type `{ folder: string }`.

KvelaGorrrrnio (changes requested):
Changes looks fine, there are a few things that needs to be changed :)

src/server/handlers/caspar-cg-handler.ts
andr9528: Renamed to `fileChangesInterval`.

src/shared/services/redux-settings-service.ts
andr9528: I have no idea why it was imported that way. Has been fixed.

andr9528: I think i know how it happend. I had VS Code generate the function for the, which is propably why.

src/server/handlers/caspar-cg-handler.ts
andr9528: Completely by accident btw.

src/shared/services/redux-settings-service.ts
```diff
@@ -45,6 +45,55 @@ export class ReduxSettingsService {
         return settingsState.generics.outputSettings[channelIndex]
     }
 
+    public clearInvalidTargetedPaths(
+        allFiles: MediaFile[],
+        originalOutputSettings: OutputSettings,
+        mediaState: Media
```
KvelaGorrrrnio, line 51: When I read the type `Media` I expect it to be related to e.g. the contents of a media file or some metadata descriptor about it. I have to look at the parameter name to understand what it is that I am looking at. Perhaps the `Media` type should be renamed to `MediaState`.

KvelaGorrrrnio (approved):

src/server/handlers/caspar-cg-handler.ts
```diff
@@ -279,24 +280,38 @@ function startTimeEmitInterval() {
...
+    fileChangesInterval = setInterval(async () => {
+        if (!waitingForCasparcgResponse) {
+            waitingForCasparcgResponse = true
+            try {
+                await getFileChanges()
+                await getThumbnailChanges()
+            } finally {
+                waitingForCasparcgResponse = false
+            }
         }
```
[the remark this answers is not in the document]
andr9528: Done

src/server/handlers/caspar-cg-handler.ts
```diff
@@ -441,23 +496,15 @@ async function loadThumbnailImage(
...
-                state.media,
+        .forEach((output: Output, channelIndex: number) => {
+            const outputFolder = new ReduxSettingsService().getOutputSettings(
+                state.settings,
                 channelIndex
-            ).thumbnailList
-            if (!isDeepCompareEqual(outputThumbnailList, outputMedia)) {
+            ).folder
+            const outputMedia = thumbnails.filter((thumbnail: ThumbnailFile) =>
+                isFolderNameEqual(thumbnail.name, outputFolder)
```
RasmusAlbrektsen, line 505: This might just be a personal opinion, but from the name `isFolderNameEqual`, i'm not entirely sure what this does. Is this having two equal folder names or that the thumbnail name equals the folder name? 
I couldn't trace the method over github for some reason.

RasmusAlbrektsen (approved):
Just one question, looks good :-)

src/server/handlers/caspar-cg-handler.ts
andr9528: The `thumbnail.name` is the full path of the file that the thumbnail is for, with CasparCG's media folder as the base.
The `outputFolder` is the folder we want thumbnails for.
The function checks if the thumbnail is located in the folder we want thumbnails for.

Hope this clarifies it. Not sure if the function needs a new name.

RasmusAlbrektsen: I think my confusions stems from the fact that `thumbnail.name` is actually a path to it's file, at not the actual file name/thumbnail name.

PR #92, opened 2023-08-28
fix: fixes overlays being applied at startup when not suppose to.

andr9528:
Adds check for if overlay should be applied, instead of always playing overlay.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2,368 +2,377 @@
 # yarn lockfile v1
 
 
-"@ampproject/remapping@^2.1.0":
-  version "2.2.0"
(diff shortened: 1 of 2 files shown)


yarn.lock
RasmusAlbrektsen, line 1: Are all these updates intentional? Just making sure :-)

andr9528: At some point prior to making the changes for this, i must have run a `yarn install`, which then updated all the versions. Don't think it hurts.

KvelaGorrrrnio (approved):

RasmusAlbrektsen (approved):

LindvedKrvang (approved):

PR #93, opened 2023-08-28
fix: adds check to abort playing/loading file with banned characters.

andr9528:
style: applies prettier styling to some files.
fix: adds banned characters settings field.

UT-252


KvelaGorrrrnio (commented):
This seems like a bandaid fix. Checking for "banned" characters do not solve the root problem. If another problematic path or some other command send to CasparCG fails, then Cliptool will still crash. Instead you should look at where the crash is caused and add proper error handling.

src/client/shared/services/socket-play-service.ts
```diff
@@ -7,11 +8,34 @@ export class SocketPlayService {
...
-    public loadFile(outputIndex: number, fileName: string) {
+    public loadFile(outputIndex: number, fileName: string): void {
+        if (this.checkIfContainsBannedCharacters(fileName)) {
+            alert('Attempted to load file with banned character. Aborting')
+            return
+        }
         this.socket.emit(ClientToServerCommand.PGM_LOAD, outputIndex, fileName)
     }
+
+    // Can very likely be done using regex - However I don't have the experience with regex to know how.
```
LindvedKrvang, line 27: Comments like these provides no value and should be removed.
Comments should explain _why_ does the code do as it does.
This comment should have been a dialogue where you asked someone for input on your approach.

Here is a good website where you can mess around with regex. It also has some descriptions as to what the differnt rules does:
https://regexr.com/

src/client/shared/services/socket-play-service.ts
```diff
@@ -7,11 +8,34 @@ export class SocketPlayService {
...
+    private checkIfContainsBannedCharacters(fileName: string): boolean {
+        const bannedCharacters =
+            state.settings.generics.ccgSettings.bannedCharacters.toLowerCase()
```
LindvedKrvang, line 30: When saving the "banned" characters you could save them directly as lowercase (or uppercase, whatever is preferred) and then you don't have to ensure the casing every time you need to use them.

src/client/shared/services/socket-play-service.ts
```diff
@@ -7,11 +8,34 @@ export class SocketPlayService {
...
+        const loweredFileName = fileName.toLowerCase()
+
+        for (let i = 0; i < bannedCharacters.length; i++) {
+            const character = bannedCharacters.charAt(i)
+            if (loweredFileName.includes(character)) {
+                return true
+            }
+        }
+        return false
+    }
```
LindvedKrvang, line 40: Here is an implementation that does not use regex, but instead utilizes the funtions of `Array`:
```
function doesStringContainBannedCharacter(stringToCheck: string, bannedCharacters: string[]): boolean {
  return stringToCheck
    .toLowerCase()
    .split('')
    .some(char => bannedCharacters.includes(char))
}
```
This test:
image

Gives the following output:
image

This is mainly to give you an idea of how you can utilize the functions of `Array` instead of using normal loops :)

LindvedKrvang (approved):
I agree with Anders that this seems like a bandaid solution.
However, I'm not the PO of this project nor do I know what features the PO wants. (But it might be an idea to propose a solution closer to the core of the problem)

olzzon:
Regarding banned characters.
One of the Main weakness in ClipTool is stil that the ACMP & OSC protocols are not normalized when received.
Fixing checks for upper, lower case, slash backslash, errors with foldernames and filenames etc.

This is still (IMHO) the most important thing for the codebase.

andr9528:
From everybodies input, i think i am going to try and add some error handling around the playing/loading of files.

Not entirely sure with the current setup, how i'd notify the user about the failure.
Don't think a log statement in the console is sufficient in this case, but should propably also be there.
Could use some input / chat about that. (@KvelaGorrrrnio & @olzzon)

One benefit of how i did it at the moment, is that it happens clientside, so i can easily show an alert.

andr9528:
Changes in this PR might be discarded, depending on where the error it is to solve, has been spottet.
Locally, with a newer version (2.4.0-NRK-RC-3eaf76) of CasparCG than other places (afaik), the issue isn't present.

RasmusAlbrektsen:
Sounds like this is a WIP again, might this benefit from being turned into a draft PR to reflect that state?

andr9528:
Due to the error explained by Kevin not being present in either my version of CasparCG and the version run in Staging, i have cleanred up the code such that only general improved code remains.

The changes should all now just be generally code improvements, rather than fixes to an issue.

@RasmusAlbrektsen Yes and No. 
It should propably have been made a draft, ones i noticed that the issue it was to fix, wasn't an issue friday.
But now that i have removed the changes relevant to solving the issue (adding check for banned characters), it should be reviewed and hopefully merged really quickly.

src/client/shared/services/socket-play-service.ts
andr9528: Banned characters are gone

src/client/shared/services/socket-play-service.ts
andr9528: Banned characters are gone

src/client/shared/services/socket-play-service.ts
andr9528: Banned characters are gone

src/server/services/casparcg-playout-service.ts
```diff
@@ -0,0 +1,156 @@
...
+        layerIndex: number,
+        fileName: string
+    ): Promise<void> {
+        if (!fileName) {
+            return
+        }
+        await this.scale(channelIndex, layerIndex)
+
+        await this.casparCgHandlerService
+            .getCasparCgConnection()
```
KvelaGorrrrnio, line 104: Is it necessary to get the connection multiple times inside the same scope?
And design-wise: Is it necessary to get the connection out? Couldn't e.g. `loadHtmlPage` be directly on `casparCgHandlerService`?

andr9528: During the work for UT-254, i had already changed the CasparCG connection to be supplied as an input to the constructor, to get rid of a circular dependency. Have applied said change here too now.

src/server/handlers/caspar-cg-handler.ts
```diff
@@ -224,7 +227,11 @@ function loadInitialOverlay(): void {
             index
         )
         if (outputSettings.webState) {
-            playOverlay(index, 10, outputSettings.webUrl)
+            casparCgPlayoutService
+                .playOverlay(index, 10, outputSettings.webUrl)
+                .then(() => {
+                    logger.info(`Loaded initial overlay for channel: ${index}`)
+                })
```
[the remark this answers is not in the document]
andr9528: Gone

src/client/shared/observers/connection-observer.ts
```diff
@@ -18,5 +18,16 @@ export class ConnectionObserver {
...
+            const splitError = error.split('|')
+            if (splitError.length !== 2) {
+                console.log(
+                    'Received error from backend without expected layout:',
+                    error
+                )
+                return
+            }
+            alert(splitError[0])
+        })
```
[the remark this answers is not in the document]
andr9528: In the `SocketIOServerHandlerService` there is a method that if used to send error events (`notifyAboutError`). That method sends a message i define, and the errors message as one string split by the `|`.

This was acutally something i'd like to have a chat about how to best notify the client that an error as occured. In the future (after UT-203), it is going to be much easier, as many opperations are simply rest calls, that can then return different HTTP responses.

src/server/services/casparcg-playout-service.ts
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    public async playOverlay(
+        channelIndex: number,
+        layerIndex: number,
+        fileName: string
+    ): Promise<void> {
+        if (!fileName) {
+            return
+        }
```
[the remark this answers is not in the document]
andr9528: Don't entirely know what we'd rather have. If i change this to throw an error, then usage sites needs to add catches to their `playOverlay(...).then(...)`, or risk getting unhandled rejecion errors. In those catches as i see it there wouldn't really be anything to do, other than maybe log it.

Would like a chat about how best to do this.

src/server/services/casparcg-playout-service.ts
```diff
@@ -0,0 +1,156 @@
...
+                    channelIndex + 1,
+                    layerIndex + 1,
+                    fileName,
+                    loopState,
+                    transitionType,
+                    transitionTime
+                )
+            return Promise.resolve()
+        } catch (error) {
+            return Promise.reject(error)
```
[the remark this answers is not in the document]
andr9528: You are right. Have removed that, as the error is caught at usage sites.

src/server/services/casparcg-playout-service.ts
```diff
@@ -0,0 +1,156 @@
...
+                .getCasparCgConnection()
+                .play(
+                    channelIndex + 1,
+                    layerIndex + 1,
+                    fileName,
+                    loopState,
+                    transitionType,
+                    transitionTime
+                )
+            return Promise.resolve()
```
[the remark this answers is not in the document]
andr9528: Gone

src/server/services/casparcg-playout-service.ts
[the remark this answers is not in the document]
KvelaGorrrrnio: I think it is a fine solution to log when e.g. playing an overlay fails. Either as an error or a warning.

src/client/shared/observers/connection-observer.ts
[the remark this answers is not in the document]
KvelaGorrrrnio: This syntax seems very implicit. If the `error` event is something we have control over, then it would be better to use a structured data type, e.g. send something like:
```typescript
interface ErrorEvent {
  message: string
  shouldNotify: boolean
}
```

src/server/handlers/caspar-cg-handler.ts
```diff
@@ -182,9 +182,19 @@ function reinvigorateChannel(
...
+                logger.info(
+                    `Re-loaded ${cuedFileName} on channel index ${index}.`
+                )
+            })
+            .catch((error) => {
+                socketIoServerHandlerService.notifyAboutError(
+                    `Caught failed attempt to load media during re-invigoration of channel ${index}.`,
+                    error
+                )
+            })
```
KvelaGorrrrnio, line 197: The curly-brackets are unnecessary in both then and catch.

src/server/services/express-service.ts
```diff
@@ -19,10 +20,14 @@ export class ExpressService {
 
         this.app.use('/', express.static(path.join(__dirname, '../../client')))
 
+        this.configureOnEvents()
```
KvelaGorrrrnio, line 23: If you return the `SocketIOServerHandlerService` from the `configureOnEvents` method, you can assign it to the `socket.IoServerHandlerService` in the constructor. This way you don't need to force the type of the attribute and can you can keep the `readonly` access keyword.

KvelaGorrrrnio (approved):
I approve under the assumption that the following comments are addressed :)

src/client/shared/observers/connection-observer.ts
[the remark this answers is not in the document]
andr9528: Before the pipe i had a human readable text, that could be displayed to a user describing the error.
Have changed the event firing method to give a `ErrorEvent` similar to the suggested.

src/server/handlers/caspar-cg-handler.ts
andr9528: Gone

src/server/services/express-service.ts
andr9528: The constructor for `SocketIOServerHandlerService` needs a `CasparCG` instance, which can't be supplied in the constructor, due to the `ExpressService` being Singleton.

RasmusAlbrektsen (approved):

src/server/services/casparcg-playout-service.ts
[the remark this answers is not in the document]
andr9528: Have changed it to throw error, and log it when caught.

PR #95, opened 2023-09-11
Ut 234/update state on reconnect

andr9528:
- Updates setings of Cliptool on startup if currently playing/loaded files on CasparCG doesn't match what Cliptool knows about.
- Resends play/load command to CasparCG when Cliptool gains connection to it, and there isn't anything being played/loaded.


KvelaGorrrrnio (approved):

src/server/services/casparcg-handler-service.ts
```diff
@@ -158,22 +219,78 @@ export class CasparCgHandlerService {
...
+            return outputSettings
+        }
+        logger.debug(
+            `File playing or loaded on CasparCG channel ${index} is different than saved file. Updating saved value: ${fileName}`
+        )
+        if (!isPaused) {
+            outputSettings.selectedFileName = fileName
+        } else if (isPaused) {
+            outputSettings.cuedFileName = fileName
+        }
```
LindvedKrvang, line 289: `isPaused` is a boolean, so no need for the `else if` as it will always be true when we reach it.
Just do:
```
if (!isPaused) {
  ...
} else {
  ...
}
```

src/server/services/casparcg-playout-service.ts
```diff
@@ -2,17 +2,42 @@ import { ReduxSettingsService } from '../../shared/services/redux-settings-servi
 import { state } from '../../shared/store'
 import { Enum as CcgEnum } from 'casparcg-connection/dist/lib/ServerStateEnum'
 import { CasparCG } from 'casparcg-connection'
+import { logger } from '../utils/logger'
+import { ErrorEvent } from '../../shared/models/error-models'
 
+/* Todo: Fix playing a file while connection to CasparCG isn't there waiting indefinitely, and then fires on reconnect.
+         Should probably just time out after e.g 10 seconds.
+*/
```
LindvedKrvang, line 10: Todos like this should be tasks in Jira instead of comments in code

src/server/services/casparcg-playout-service.ts
```diff
@@ -2,17 +2,42 @@ import { ReduxSettingsService } from '../../shared/services/redux-settings-servi
...
 export class CasparCgPlayoutService {
     private readonly reduxSettingsService: ReduxSettingsService
     private readonly casparCgConnection: CasparCG
+    private readonly socketServer: any
```
LindvedKrvang, line 14: You don't have a type for the SocketServer?

LindvedKrvang (approved):

src/server/services/casparcg-playout-service.ts
andr9528: Nice catch. Missed adding type to it, while doing so for the PR of UT-254. Will be added in a moment.

src/server/services/casparcg-playout-service.ts
andr9528: Created bug task UT-257 from todo, and removed todo.

src/server/services/casparcg-handler-service.ts
andr9528: Updated.

PR #96, opened 2023-09-12
fix: fixes packaged Cliptool-gateway being unable to run.

andr9528:
UT-250


RasmusAlbrektsen (approved):

KvelaGorrrrnio (approved):

KvelaGorrrrnio:
I still think it is weird that only the gateway fails, since they are both using `pkg` for packaging.

andr9528:
Yes, even more so that the files that import the offending package (@tv2/logger/node) are 100% identical in the gateway and server part of the code. Where the server one works, but the gateway one didn't.

PR #102, opened 2023-11-28
Small bugfixes

andr9528:
Found and fixes 3 small bugs while fixing staging PC's Cliptool/CasparCG not communicating correctly.

- Fixes attempt to mutate redux state
- Fixes not using DefaultLayer for the layer to play/load files and webpages on.
- Fixes output settings being cleared on startup, when CasparCG is not playing anything. 

Take note of the base branch, as that was the branch i branched out from.


LindvedKrvang (approved):
Take this approve with a grain a salt since I don't know what your changes are doing.
You have removed a parameter a lot of places and replaced it with a "fixed" value from settings.
I can't tell if that's a good or a bad thing.

PR #104, opened 2023-12-05
Apply feedback from Kasper

andr9528:
- Fixes overlay playing on the same layer as mediafiles, when they should play on the layer just above (playout layer + 1), according to Kasper.
  - This also explains why they were given a '10' for the layer, while playing was given a '9' before i fixed it not using the value from settings.
- Adds some more images to a couble of sections in the Readme.
- Updated some images to include/use `Cam 2` as the example, per request of Kasper :)
- Renamed a method argument to better match content.

Kasper also wanted the section order of the readme to be inverted, to be as i had originally made it. 
I have not done that - yet - as i think that needs a small talk in the team, and possibly Jacobs input on what he finds most suitable


README.md
```diff
@@ -43,7 +45,7 @@ That section should be moved to be part of the configuration, and of course, un-
 
 In case the config file has removed the `osc` section for some reason, then it can also be found below.
 
-The port defined under the `predefined-client` part of the `osc` section should match the port inputted to the [CasparCG Settings](#casparcg-settings)
+The port defined under the `predefined-client` part of the `osc` section should match the port inputted to the [casparCG settings](#casparcg-settings).
```
KvelaGorrrrnio, line 48: Is it intentional that the first `c` in `CasparCG` is lowercased?

KvelaGorrrrnio (approved):

README.md
andr9528: Huh you are right. As that is a name, it should remain capatilized.

andr9528: Fixed.

LindvedKrvang (approved):

olzzon (approved):

PR #106, opened 2024-01-15
Feat/refactor outputsettings outputstate

olzzon:
There was no initial default state of loop, mix, manual, overlay. So the settings in the settings page was just a copy of the current state.
Initial state has been added to outputSettings, and outputSettings has been refactored and called outputState.

### DIFF
--- a/src/client/settings-page/components/checkbox/checkbox.tsx
+++ b/src/client/settings-page/components/checkbox/checkbox.tsx
@@ -13,7 +13,7 @@ export default function Checkbox(props: CheckboxProps): JSX.Element {
   return (
     <div className={`c-checkbox ${props.className ?? ''}`}>
       <label>
-        {props.description ? props.description : ''}
+        {props.description ?? ''}
         <input type="checkbox" 
           checked={props.checked} 
           onChange={props.onChange}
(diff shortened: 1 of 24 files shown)


src/server/services/hidden-files-persistence-service.ts
```diff
@@ -101,10 +101,10 @@ export class HiddenFilesPersistenceService {
     }
 
     private hasSelectedOrCuedHiddenFiles(hiddenFiles: HiddenFiles): boolean {
-        const outputs: OutputSettings[] =
+        const outputs: OutputState[] =
             this.reduxSettingsService.getGenericSettings(
                 state.settings
-            ).outputSettings
+            ).outputsState
```
KvelaGorrrrnio, line 107: To indicate that it is a collection of states for different outputs, I would suggest to rename `outputsState` to `outputStates`

KvelaGorrrrnio (approved):

src/client/settings-page/components/checkbox/checkbox.tsx
```diff
@@ -12,15 +12,13 @@ interface CheckboxProps {
...
-        <label className="label">
-          {props.description}
-        </label>
-      ) : ''}
-      <input type="checkbox" 
-        checked={props.checked} 
-        onChange={props.onChange}
-        className='c-checkbox__input' />
+      <label>
+        {props.description ? props.description : ''}
```
LindvedKrvang, line 16: I think you can write this as `props.description ?? ''` if you want

LindvedKrvang (approved):

src/client/settings-page/components/checkbox/checkbox.tsx
olzzon: Agree, that looks way better :)

PR #109, opened 2024-01-19
Feat/k2 emulator amp merge

olzzon:
Improved AMP protocol with support for GV mixer.
This is the code from https://github.com/olzzon/casparcg-cliptool/tree/feat/k2-emul implemented into the TV2 ClipTool rewrite.
This code needs to be tested further in real life before going into production, both on Ross and on GV installations.


LindvedKrvang (approved):

src/gateway/gateways/AMP/utils.ts
```diff
@@ -0,0 +1,115 @@
...
+    response += status.playing ? '81' : 'a0'
+    response += status.playing ? '80' : '81'
+    response += status.automode ? '81' : '01'
+    response += '400020000000000000'
+    let cueStart = status.cueing ? 'a2' : '62'
+    if (status.playing) cueStart = '22'
+    response += cueStart
+    response += '00'
+
+    response = '7f2002a08181400020000000000000a200'
```
KvelaGorrrrnio, line 89: It would be nice if the string literals here were extracted to properly named constants. You can infer some of the informations, but I have no idea why we are checking `status.playing` twice or what `400020000000000000` is supposed to mean :)

src/gateway/gateways/AMP/utils.ts
```diff
@@ -0,0 +1,115 @@
...
+    workingFolder: string
+): string[] {
+    let ampFileList: string[] = mediaFiles
+        .filter((file) => {
+            return file.name.startsWith(workingFolder)
+        })
+        .map((file) => {
+            return getFileNameNoPath(file.name)
+        })
+    return ampFileList
```
KvelaGorrrrnio, line 110: The local variable can be refactored away.
```typescript
    return mediaFiles
        .filter((file) => file.name.startsWith(workingFolder))
        .map((file) => getFileNameNoPath(file.name))
```

KvelaGorrrrnio (approved):

src/gateway/gateways/AMP/utils.ts
olzzon: I agree. :)

